PR #1, opened 2018-01-11
Create LICENSE

Hanzei:


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 Yusuke Nemoto
```
kaakaa: Would you like to add your name?

Hanzei: I dont want to add my real name. Do you know if its possible to add a pseudonym?

kaakaa: I understood your thoughts.

My suggestion of previous comment did not concern the legal problem, I just wanted to add your name as a collaborative developer.
I think that It's possible to write pseudonym as copyright holder, but it may not have no meaning.
And for many country's law, without real name the copyright of code is protected by [Berne Convention](https://en.wikipedia.org/wiki/Berne_Convention)

I will respect your will as to whether to include your name.
Thanks.

Hanzei: Thank you for your kind words. I added my pseudonym to the license.

PR #11, opened 2018-01-23
Fix bugs

Hanzei:
This PR fixes a couple of bugs:
- We now store `plugin.API` to perform action with it. This will be needed in other PRs.
- If the user provides less then two arguments (e.g. `/matterpoll "Question" "Answer 1"`, a help message will be posted. This PR also adds tests for this chase.
- The help test is now properly formated.

### DIFF
--- a/plugin.go
+++ b/plugin.go
@@ -24,7 +24,7 @@ func (p *MatterpollPlugin) OnActivate(api plugin.API) error {
 
 func (p *MatterpollPlugin) ExecuteCommand(args *model.CommandArgs) (*model.CommandResponse, *model.AppError) {
 	input := ParseInput(args.Command)
-	if len(input) < 3 {
+	if len(input) < 2 {
 		return &model.CommandResponse{
 			ResponseType: model.COMMAND_RESPONSE_TYPE_EPHEMERAL,
 			Username:     `Matterpoll`,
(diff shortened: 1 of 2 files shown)


plugin.go
```diff
@@ -22,11 +24,11 @@ func (p *MatterpollPlugin) OnActivate(api plugin.API) error {
 
 func (p *MatterpollPlugin) ExecuteCommand(args *model.CommandArgs) (*model.CommandResponse, *model.AppError) {
 	input := ParseInput(args.Command)
-	if len(input) == 0 {
+	if len(input) < 3 {
```
kaakaa: I suppose that there may be a poll having just one answer.
e.g.) "Let's have a lunch together" "Yes!"

So I think that the number here had better to be `2`, not `3`,

plugin_test.go
```diff
@@ -37,31 +37,55 @@ func TestPluginExecuteCommand(t *testing.T) {
...
+	p := &main.MatterpollPlugin{}
+
+	r, err := p.ExecuteCommand(&model.CommandArgs{
+		Command: `/matterpoll "abcd" "abcd"`,
+	})
+	assert.Nil(t, err)
+	isHelpResponce(t, r)
+}
+
+func isHelpResponce(t *testing.T, r *model.CommandResponse) {
```
kaakaa: This assert function name should be `assertHelpResponse` since I think that functions started with `is`  or `has` should return true/false.

Hanzei: :+1:

plugin.go
Hanzei: :+1:

Hanzei:
@kaakaa Changes made

PR #15, opened 2018-08-20
Initial Version

Hanzei:
This is a first working version.

Please go ahead and test it. Especially I would really like to get feedback on the ux.


server/poll.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/mattermost/mattermost-server/model"
+)
+
+type Poll struct {
+	Creator  string
+	Question string
+	Options  []*Option
+}
+
+type Option struct {
```
kaakaa: Similarly, I think that `Option` should have `(option)ID`.
Considering Poll as data, I think that it would be better if we could identify Option one by one.

server/poll.go
```diff
@@ -0,0 +1,147 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/mattermost/mattermost-server/model"
+)
+
+type Poll struct {
```
kaakaa: I want to add `DataSchemaVersion` (e.g.: "V1") to `Poll` for the future.

Additionally, I think that `Poll` should have `PollID` filed, because I think that it is preferable for data to be self-describing. If DataSchemaVersion should be changed, It is easier to write function.

How do you think about this?

server/handler.go
```diff
@@ -0,0 +1,190 @@
...
+		return
+	}
+
+	if userID != poll.Creator {
+		response.EphemeralText = endPollInvalidPermission
+		writePostActionIntegrationResponse(w, response)
+		return
+	}
+
+	appErr = p.API.KVDelete(pollID)
```
kaakaa: If error happens after here, poll data is disappear and can't be recovered.
So I think `KVDelete` should be run the end of this function (before writePostActionIntegrationResponse).

Hanzei: Good catch! I will fix this.

server/poll.go
Hanzei: I like the idea of a version. I will add it to `poll` and the api.

Hanzei:
Hey @kaakaa,
Thank you very much for your feedback. Lets merge this one for now and discuss the ids for polls and options later.

PR #24, opened 2018-08-23
Add config option for trigger

Hanzei:
Todo:
- [x] Add tests
- [x] Fix tests

Fixes #16

PR #33, opened 2018-08-24
Show how many pepole have voted

### LINKED ISSUE #27: Display how many people have voted
We cannot know how many people have voted until poll is ended.
It had better to display the number of votes on poll post

Sample:

Hanzei:
There was actually some needed to get this working because `handleVote()` didn't updated the post until now.

I'm not 100% happy with adding `SiteURL` to `MatterpollPlugin` but I think its okay for now. We can work on this after #24 is merged.

i'm 2/5 on the text. If you have a better idea, go ahead and change it. 

Fixes #27

### DIFF
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -25,6 +25,7 @@ func (p *MatterpollPlugin) OnConfigurationChange() error {
 		return err
 	}
 
+	p.ServerConfig = p.API.GetConfig()
 	p.Config = c
 	return nil
 }
--- a/server/handler.go
+++ b/server/handler.go
@@ -75,9 +75,9 @@ func (p *MatterpollPlugin) handleVote(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	post := model.Post{}
-	post.AddProp("attachments", poll.ToPostActions(p.SiteURL, pollID, displayName))
-	response.Update = &post
+	post := &model.Post{}
+	post.AddProp("attachments", poll.ToPostActions(*p.ServerConfig.ServiceSettings.SiteURL, pollID, displayName))
+	response.Update = post
 
 	if hasVoted {
 		response.EphemeralText = voteUpdated
--- a/server/handler.go
+++ b/server/handler.go
@@ -53,6 +53,13 @@ func (p *MatterpollPlugin) handleVote(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	displayName, appErr := p.ConvertUserToDisplayName(poll.Creator)
+	if appErr != nil {
+		response.EphemeralText = commandGenericError
+		writePostActionIntegrationResponse(w, response)
+		return
+	}
+
 	hasVoted := poll.HasVoted(userID)
 	err := poll.UpdateVote(userID, optionNumber)
 	if err != nil {
@@ -60,12 +67,17 @@ func (p *MatterpollPlugin) handleVote(w http.ResponseWriter, r *http.Request) {
 		writePostActionIntegrationResponse(w, response)
 		return
 	}
+
 	appErr = p.API.KVSet(pollID, poll.Encode())
 	if appErr != nil {
 		response.EphemeralText = commandGenericError
 		writePostActionIntegrationResponse(w, response)
 		return
 	}
+
+	post := model.Post{}
+	post.AddProp("attachments", poll.ToPostActions(p.SiteURL, pollID, displayName))
+	response.Update = &post
 
 	if hasVoted {
 		response.EphemeralText = voteUpdated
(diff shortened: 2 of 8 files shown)


server/poll.go
```diff
@@ -53,23 +56,32 @@ func (p *Poll) ToCommandResponse(siteURL, authorName, pollID string) *model.Comm
 		},
 	})
 
-	return getCommandResponse(model.COMMAND_RESPONSE_TYPE_IN_CHANNEL, "", []*model.SlackAttachment{{
+	var hasText string
+	if numberOfVotes == 1 {
+		hasText = "has"
+	} else {
+		hasText = "have"
+	}
```
Hanzei: This looks ugly. @kaakaa do you have a idea how improve this?

server/handler.go
```diff
@@ -45,20 +45,32 @@ func (p *MatterpollPlugin) handleVote(w http.ResponseWriter, r *http.Request) {
...
 	}
+
 	appErr = p.API.KVSet(pollID, poll.Encode())
 	if appErr != nil {
 		response.EphemeralText = commandGenericError
 		writePostActionIntegrationResponse(w, response)
 		return
 	}
 
+	post := model.Post{}
```
kaakaa: If we can use `PostActionIntegrationResponse.PostId`, we can get Post instance, should change only `Post.Text`, and can avoid to add `SiteURL` to `MatterpollPlugin`.
However, even though this PR was merged, `PostActionIntegrationResponse.PostId` looks having no value now.

Do you know why can't get `PostActionIntegrationResponse.PostId`?

kaakaa: I'm on the latest master branch and developer environment.
"Delete Poll" still reply `This feature is only available on Mattermost v5.3.`.

Hanzei: Lets continue the discussion on pre-release

Hanzei:
I think rebuilding the whole post is not bad. If e.g. display names change, we also update. 

About the `SiteURL`: After having some trouble with http vs https (See https://pre-release.mattermost.com/core/pl/bc6odccxnby4mymkhr6knbdw9e), propose the following; 
- We fetch the server config with [`GetConfig()`](https://developers.mattermost.com/extend/plugins/server/reference/#API.GetConfig) and save it in `MatterpollPlugin`
- We use `ServiceSettings.SiteURL` to create polls

What do you think @kaakaa?

kaakaa:
Yes, I'm okay with rebuilding the whole post. I think it is easier to understand code.

I have two concerns about `SiteURL`.

At first, `GetConfig()` returns the pointer (i.e. `*model.Config`), and I suppose that the pointer isn't updated unless mattermose-server is restarted or matterpoll is re-enabled. If a configuration change is made that does not require a restart, we will result in keeping the pointer to old config not in use.  So if we use config object, we should call `p.GetConfig()` at every time, or we should have a SiteURL value only.

Second, I couldn't confirm `commandArgs.SiteURL` problem. I've tested it on mattermost via nginx with TLS settings, and `commandArgs.SiteURL` returned correct SiteURL value (`https://${hostname}/${subpath}`). Are you using `
Forward port 80 to 443:` settings?

Hanzei:
> At first, `GetConfig()` returns the pointer (i.e. `*model.Config`), and I suppose that the pointer isn't updated unless mattermose-server is restarted or matterpoll is re-enabled. If a configuration change is made that does not require a restart, we will result in keeping the pointer to old config not in use.  So if we use config object, we should call `p.GetConfig()` at every time, or we should have a SiteURL value only.

We already implement the [`OnConfigurationChange()`](https://developers.mattermost.com/extend/plugins/server/reference/#Hooks.OnConfigurationChange) hook. So we just have call `GetConfig()` in the hook.

> Second, I couldn't confirm `commandArgs.SiteURL` problem. I've tested it on mattermost via nginx with TLS settings, and `commandArgs.SiteURL` returned correct SiteURL value (`https://${hostname}/${subpath}`). Are you using `Forward port 80 to 443:` settings?

`Forward port 80 to 443` can only we used if you don't use a proxy. My Site is available via http and https. But even via https `commandArgs.SiteURL` always returns the http URL.

Hanzei:
Ready for the next review

server/configuration.go
```diff
@@ -25,6 +25,7 @@ func (p *MatterpollPlugin) OnConfigurationChange() error {
 		return err
 	}
 
+	p.ServerConfig = p.API.GetConfig()
```
Hanzei: I didn't check for `nil`consciously. If fetching `GetConfig()` fails, what do we do?

I'm quite sure this will never be a problem.

kaakaa: Agree. If `GetConfig()` is failed, it's not sure that other plugin functions work correctly too. Matterpoll can do nothing here.

kaakaa:
> We already implement the OnConfigurationChange() hook. So we just have call GetConfig() in the hook.

You're right! `OnConfigurationChange` can catch any update on system console, so my concern is disappear. Thank you 👍 

> Forward port 80 to 443 can only we used if you don't use a proxy. My Site is available via http and https. But even via https commandArgs.SiteURL always returns the http URL.

I see. As you say, using `ServiceSettings.SiteURL` is better than `commandArgs.SiteURL`.

PR #37, opened 2018-08-25
 Improve installation notes

Hanzei:
Lets see if this fixes the problem. If it does we can merge. See https://pre-release.mattermost.com/core/pl/wim3gnm43tr55gac98fowk33jw

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ Supported Mattermost Server Versions: 5.2+
 
 1. Go to the [releases page of this Github repository](https://github.com/matterpoll/matterpoll/releases/latest) and download the latest release for your Mattermost server.
 2. Upload this file in the Mattermost **System Console > Plugins > Management** page to install the plugin. To learn more about how to upload a plugin, [see the documentation](https://docs.mattermost.com/administration/plugins.html#plugin-uploads).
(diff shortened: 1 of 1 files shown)


Hanzei:
I removed the statement about `AllowedUntrustedInternalConnections` for now. I'm still not sure what the root cause of the problem is and how to fix it.

README.md
```diff
@@ -14,6 +14,7 @@ Supported Mattermost Server Versions: 5.2+
 
 1. Go to the [releases page of this Github repository](https://github.com/matterpoll/matterpoll/releases/latest) and download the latest release for your Mattermost server.
 2. Upload this file in the Mattermost **System Console > Plugins > Management** page to install the plugin. To learn more about how to upload a plugin, [see the documentation](https://docs.mattermost.com/administration/plugins.html#plugin-uploads).
+3. You should set `EnablePostUsernameOverride` and `EnablePostIconOverride` to `true` in your `config.json`.
```
kaakaa: I think we should avoid to edit config.json directly.

3. You should set `true` to **Enable integrations to override usernames** and **Enable integrations to override profile picture icons** in **System Console > Custom Integrations**.

How about this?

PR #38, opened 2018-08-27
Release v0.0.3

Hanzei:
Requires:
- [ ] #37


Hanzei:
I would like to also add #39 to this, if its fine for you @kaakaa

PR #45, opened 2018-08-29
Fix command error messages

kaakaa:
Matterpoll can set their trigger word.
But even though trigger word isn't `matterpoll`, error message tell `matterpoll` as trigger word.

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"fmt"
+
 	"github.com/mattermost/mattermost-server/model"
 	"github.com/mattermost/mattermost-server/plugin"
 )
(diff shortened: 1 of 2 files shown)


Hanzei:
This issue could also be resolved by using https://golang.org/pkg/text/template/. It would make for a more fexible solution, but I'm not sure if its worth the time.

kaakaa:
Yes, I know.  But I felt that using text/template package is over spec. if we use text/template, better to create new function handling all messages.
If you like, I'll work on that as another enhancement issue.

Hanzei:
Lets keep this option in mind. But for now this is fine. For two templates it would be to much work.

Lets merge this :smile:

PR #46, opened 2018-08-29
Add message notifying that poll has ended

kaakaa:
Even if poll has ended, @ mention is not triggered. So user cannot notice the poll is over. 
By this PR, when ending poll, matterpoll posts message about that poll end.

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"io"
 	"net/http"
 	"strconv"
@@ -19,8 +20,9 @@ const (
 	voteCounted = "Your vote has been counted."
 	voteUpdated = "Your vote has been updated."
 
-	endPollSuccessfully      = "Poll has ended. Let's see results!\nYou can see results as the parent post of this."
-	endPollInvalidPermission = "Only the creator of a poll is allowed to end it."
+	// Parameter: Question, Permalink
+	endPollSuccessfullyFormat = "The poll **%s** has ended.  The original post have been updated. You can jump to it by pressing [here](%s)"
+	endPollInvalidPermission  = "Only the creator of a poll is allowed to end it."
 
 	deletePollInvalidPermission = "Only the creator of a poll is allowed to delete it."
 	deletePollSuccess           = "Succefully deleted the poll."
@@ -98,6 +100,10 @@ func (p *MatterpollPlugin) handleVote(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	var teamID string
+	if v, ok := request.Context["team_id"]; ok {
+		teamID = v.(string)
+	}
 	post := &model.Post{}
 	model.ParseSlackAttachment(post, poll.ToPostActions(*p.ServerConfig.ServiceSettings.SiteURL, pollID, displayName))
 	response.Update = post
@@ -161,18 +167,40 @@ func (p *MatterpollPlugin) handleEndPoll(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	writePostActionIntegrationResponse(w, response)
+	p.postEndPollAnnouncement(request, poll.Question)
 
-	pollPost, _ := p.API.GetPost(request.PostId)
+	writePostActionIntegrationResponse(w, response)
+}
+
+func (p *MatterpollPlugin) postEndPollAnnouncement(request *model.PostActionIntegrationRequest, question string) {
+	var teamID string
+	if v, ok := request.Context["team_id"]; !ok {
+		return
+	} else {
+		teamID = v.(string)
+	}
+
+	team, err := p.API.GetTeam(teamID)
+	if err != nil {
+		return
+	}
+	permalink := fmt.Sprintf("%s/%s/pl/%s", *p.ServerConfig.ServiceSettings.SiteURL, team.Name, request.PostId)
+
+	pollPost, err := p.API.GetPost(request.PostId)
+	if err != nil {
+		return
+	}
+	channelID := pollPost.ChannelId
+
 	endPost := &model.Post{
 		UserId:    request.UserId,
-		ChannelId: pollPost.ChannelId,
+		ChannelId: channelID,
 		RootId:    request.PostId,
-		Message:   endPollSuccessfully,
+		Message:   fmt.Sprintf(endPollSuccessfullyFormat, question, permalink),
 		Type:      model.POST_DEFAULT,
 		Props: model.StringInterface{
 			"override_username": responseUsername,
-			"override_icon_url": responseIconURL,
+			"override_icon_url": fmt.Sprintf(responseIconURL, *p.ServerConfig.ServiceSettings.SiteURL, PluginId),
 			"from_webhook":      "true",
 		},
 	}
--- a/server/api.go
+++ b/server/api.go
@@ -199,7 +199,8 @@ func (p *MatterpollPlugin) postEndPollAnnouncement(request *model.PostActionInte
 			"from_webhook":      "true",
 		},
 	}
-	if _, err := p.API.CreatePost(endPost); err != nil {
+	_, err = p.API.CreatePost(endPost)
+	if err != nil {
 		p.API.LogError(endPollAnnouncementPostError, "details", "failed to CreatePost")
 	}
 }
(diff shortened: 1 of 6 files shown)


Hanzei:
Hey @kaakaa 

First of all I think this changes is needed. :+1: 

But I have to think a bit about the UX and the message which gets posted. I will review when I got the time for this.

kaakaa:
Sure!
When if you'll get an idea, please let me know.

Hanzei:
I like the feature, but I think the message needs some improvement. I see some problems:

- Users need to identify the poll right away
- "parent post" as a concept and term is never mentioned in the ui of Mattermost

I propose to use e.g. "The poll **Is Matterpoll great?** has ended. The original post have been updated. You can jump to it by pressing [here](https://example.com/)". The link would be a permalink to the original post.

kaakaa:
Thanks @Hanzei. Text you suggest looks great, and I'll work on it.

Hanzei:
But constructing the like isn't so easy. You will need to get the team name for this.

kaakaa:
Yes, you’re right.

It’s just an idea, but it may be possible to add TeamId to context of PostActionIntegration. We can get TeamId from CommandArgs when calling ExecuteCommand.

I’ll confirm its feasibility and find other ways. But it may take a few days.

Hanzei:
Sure we can do this

kaakaa:
@Hanzei I've added teamID to PostActionIntegration.Context. This PR is ready for review.

server/handler.go
```diff
@@ -15,7 +15,9 @@ const (
 	voteCounted = "Your vote has been counted."
 	voteUpdated = "Your vote has been updated."
 
-	endPollInvalidPermission = "Only the creator of a poll is allowed to end it."
+	// Parameter: Question, Permalink
+	endPollSuccessfullyFormat = "The poll **%s** has ended.  The original post have been updated. You can jump to it by pressing [here](%s)"
```
Hanzei: There is a double white space between "ended." and "The"

Hanzei:
Please also see https://pre-release.mattermost.com/core/pl/se1g1wck47raxp9ui91fwtuc4w

kaakaa:
Thank you for your reviews!
As you say, storing teamID in context is not good, and if `PostActionIntegratoinRequest` has `TeamID`  we don't need to store `TeamID` in `Poll`. Your suggestion looks perfect.
So if we can, I wait to complete [this feature](https://github.com/mattermost/mattermost-server/pull/9384) and this PR include in `v0.2.0`. Because once we store TeamID in context, we must care about version specific data. Sorry for my poor archetecture.

Of course, I'll fix other comments later. (but if adding call of logging, I must add a mocking of logging. It's ugly, so I want to think more about avoiding it)

Hanzei:
FYI https://github.com/mattermost/mattermost-server/pull/9384 just got merged. You can continue your work into this. But if we merge this, we will have to set the minimum server version to 5.4.

kaakaa:
@Hanzei  Thank you for letting me know it! I'll work on this later. And I test this feature for master branch of mattermost-server until v5.4 release.

kaakaa:
I updated this PR. This PR is almost competed, but I'll ask you to review this after merging #68.

Before asking you to review, I want your feedback about mock of logging function.
Logging function of plugin can have any number of args. `GetMockArgumentsWithType` makes it easy  to mock calling logging functions with various number of args.
https://github.com/matterpoll/matterpoll/pull/46/files#diff-2a92eec87366e63ea04281292c52f170R570

I think that this solution is not perfect, but good workaround.
How do you think about this?

Hanzei:
`GetMockArgumentsWithType` is a fine workaround. Mocking log functions is a pain and ugly. I think we are improving our workarounds.

Can we move test related fucntion like `GetMockArgumentsWithType` and `GetMockArgumentsWithType` into `plugin_test.go`? This way we keep them all together. Maybe be will find a better way to organise them like a `test` package, but that is one problem for another time.

server/api.go
```diff
@@ -160,9 +164,43 @@ func (p *MatterpollPlugin) handleEndPoll(w http.ResponseWriter, r *http.Request)
...
+		RootId:    request.PostId,
+		Message:   fmt.Sprintf(endPollSuccessfullyFormat, question, permalink),
+		Type:      model.POST_DEFAULT,
+		Props: model.StringInterface{
+			"override_username": responseUsername,
+			"override_icon_url": fmt.Sprintf(responseIconURL, *p.ServerConfig.ServiceSettings.SiteURL, PluginId),
+			"from_webhook":      "true",
+		},
+	}
+	if _, err := p.API.CreatePost(endPost); err != nil {
```
Hanzei: Nit pick: Can we stay to our usual pattern here and exclude `_, err := p.API.CreatePost(endPost)` for if

server/api_test.go
```diff
@@ -382,6 +385,77 @@ func TestHandleEndPoll(t *testing.T) {
...
+			API:      api4,
+			Request:  &model.PostActionIntegrationRequest{UserId: "userID1", PostId: "postID1", TeamId: "teamID1"},
+			Question: "Question1",
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			p := &MatterpollPlugin{
+				ServerConfig: &model.Config{ServiceSettings: model.ServiceSettings{SiteURL: model.NewString("http://localhost:8065")}},
+			}
+			p.SetAPI(test.API)
```
Hanzei: I propose to use `setupTestPlugin` here. Its used in every other test and it makes things consistent.

server/api_test.go
```diff
@@ -382,6 +385,77 @@ func TestHandleEndPoll(t *testing.T) {
...
+	api4.On("GetTeam", "teamID1").Return(&model.Team{Name: "team1"}, nil)
+	api4.On("GetPost", "postID1").Return(&model.Post{ChannelId: "channelID1"}, nil)
+	api4.On("CreatePost", mock.AnythingOfType("*model.Post")).Return(nil, &model.AppError{})
+	api4.On("LogError", GetMockArgumentsWithType("string", 3)...).Return(nil)
+	defer api4.AssertExpectations(t)
+
+	for name, test := range map[string]struct {
+		API      *plugintest.API
+		Request  *model.PostActionIntegrationRequest
+		Question string
```
Hanzei: Nit pick: Is this really needed?

server/api_test.go
```diff
@@ -465,7 +539,7 @@ func TestHandleDeletePoll(t *testing.T) {
 			assert := assert.New(t)
 
 			p := setupTestPlugin(t, test.API, samplesiteURL)
-			AllowRequestLogging(test.API)
+			test.API.On("LogDebug", GetMockArgumentsWithType("string", 7)...).Return()
```
Hanzei: :+1:

server/api_test.go
```diff
@@ -490,7 +564,3 @@ func TestHandleDeletePoll(t *testing.T) {
 		})
 	}
 }
-
-func AllowRequestLogging(api *plugintest.API) {
-	api.On("LogDebug", mock.AnythingOfType("string"), mock.AnythingOfType("string"), mock.AnythingOfType("string"), mock.AnythingOfType("string"), mock.AnythingOfType("string"), mock.AnythingOfType("string"), mock.AnythingOfType("string")).Return()
-}
```
Hanzei: :+1:

server/api.go
```diff
@@ -160,9 +164,43 @@ func (p *MatterpollPlugin) handleEndPoll(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	p.postEndPollAnnouncement(request, poll.Question)
```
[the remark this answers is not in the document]
kaakaa: 👍

server/api_test.go
kaakaa: :+1:

server/api.go
kaakaa: 👍 
Are you saying this fix, right?
https://github.com/matterpoll/matterpoll/pull/46/commits/805f4a62322d9398264b77d83011abbd853d756c

server/api_test.go
kaakaa: 👍

kaakaa:
Thanks @Hanzei!
I appreciate your reviews and suggestions.

PR #47, opened 2018-09-02
Add CreatedAt field to Poll

### LINKED ISSUE #43: Add create date to poll
We might want implement some kind of garbage collector for old polls some time. To prepare for this I propose we add a `CreatedAt` field to polls. 

Not that is wont involve any front end changes.

Hanzei:
This PR add `CreatedAt int64` to `Poll`. I chose `int64`because its used my Mattermost.

Testing this was a lit hard. I'm using https://github.com/bouk/monkey to replace the calls to `GetMillis()` by a static number. This ain't pretty, but it does the job. Otherwise we would have to do something like adding `TimeGenerator` to `MatterpollPlugin`.

I also fixed some Gopkg problems.

Fixes #43

### DIFF
--- a/server/poll.go
+++ b/server/poll.go
@@ -9,6 +9,7 @@ import (
 )
 
 type Poll struct {
+	CreatedAt         int64
 	Creator           string
 	DataSchemaVersion string
 	Question          string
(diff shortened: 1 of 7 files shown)


server/poll.go
```diff
@@ -9,6 +9,7 @@ import (
 )
 
 type Poll struct {
+	CreatedAt         int64
 	Creator           string
 	DataSchemaVersion string
```
kaakaa: Should we update DataSchemaVersion?

Hanzei: I think we should. Units v1.0.0 we can do every change we like. After v1.0.0 I will care about backwards compatibility. See https://semver.org/#spec-item-4

PR #51, opened 2018-09-07
Add poll settings

### LINKED ISSUE #35: Allow users to select options
It would be handy if the user could pick options for a poll. I could this of the following ideas:
- secret: Don't show who voted for what at the end
- progress: During the poll, show how many votes each answer option got

I'm not sure about the interface. I could think about something like `/matterpoll "Question" --secret`

hanzei:
This PR adds to two settings for polls:

- secret: Don't show who voted for what at the end
- progress: During the poll, show how many votes each answer option got

This is still quite early in development, but I want to give the possibility for feedback.
I'm 1/5 on the naming of the setting and the term "setting".

Settings parsing is quite bad, but i failed to find a working regex for this, see code comments.

Fixes #35


kaakaa:
Thank you for sharing your early implements! I left my feedbacks.

> secret: Don't show who voted for what at the end

I feel that `secret` has many meanings, so `anonymous` might be better than `secret`.

> I'm 1/5 on the naming of the setting and the term "setting".

Yes, I feel the same, and it should be the term `PollSetting`.
The others are `PollOption`, `PollKind`, `PollFeature`, or `PollConfig`... but these terms is not better than `settings`. I feel that `PollOption` is good, but it conflicts with `AnswerOption`...

> Settings parsing is quite bad, but i failed to find a working regex for this, see code comments.

I want to avoid using complex regex.
We might be able to use `flag.FlagSet` for parsing settings [(Sample Code)](https://play.golang.org/p/mvdr5fycu2G). But I'm not sure that `flag.FlagSet` is perfect. This is just a early feedback.

Hanzei:
Thank you very much for the feedback!

> I feel that secret has many meanings, so anonymous might be better than secret.

:+1: Renamed it!

> Yes, I feel the same, and it should be the term PollSetting.

:+1: Renamed it!

> I want to avoid using complex regex.

Why?

> We might be able to use `flag.FlagSet` for parsing settings [(Sample Code)](https://play.golang.org/p/mvdr5fycu2G). But I'm not sure that `flag.FlagSet` is perfect. This is just a early feedback.

The problem is that we still need something like `strings.Split()` to cut the input in pieces. But by doing so we distinguish between a setting in a question e.g. `/poll "Question -secret"` from a setting outside a question, e.g. `/poll "Question" -secret`

kaakaa:
> Why?

I thisn that regex is hard to understand completely and it may require many test cases.
So if we can achieve it by other way, I wanna avoid regex. Of course, regex is not bad.

> The problem is that we still need something like strings.Split() to cut the input in pieces. But by doing so we distinguish between a setting in a question e.g. /poll "Question -secret" from a setting outside a question, e.g. /poll "Question" -secret

Oh...It was. We can avoid that by replacing sentence surrounded by quotes with dummy word without spaces, but it's achieved by regex... `¯\_(ツ)_/¯`
Does your implementation solve this problem?

Hanzei:
I will put this feature aside for now and focus more on getting the plugin stable.

Hanzei:
After thinking a bit about this I propose to merge the "backend" part of this PR.

I would like to get this plugin stable as soon as possible. Because of this changes to `Poll` should be made until we release v1.0. I'm quite happy with the backend code and think its save to merge it. I would branch of the frontend code and use it later, when we are ready.

kaakaa:
Sure. Merging only backend code has no problem, and it's good solution for us 👍

Hanzei:
Sometime I just have to wrap my head around a problem until a solutions comes up. I end up rewriting `ParseInput()` to get this working.

The implementation is very "liberal". E.g. the following inputs are considered valid:
- `/poll question`
- `/poll "question`
- `/poll question"`
- `/poll question" --secret`
But I think its fine to be liberal here.

One change made is, that `/poll`, /poll "help"` and `/poll help` will all show the help message.

server/command.go
```diff
@@ -13,32 +13,41 @@ const (
...
-	// Parameter: Trigger, Trigger
-	commandInputErrorFormat = "Invalid Input. Try `/%s \"Question\"` or `/%s \"Question\" \"Answer 1\" \"Answer 2\" \"Answer 3\"`"
+	commandHelpTextFormat = "To create a poll with the answer options \"Yes\" and \"No\" type `/%[1]s \"Question\"`.\n" +
+		"You can customise the options by typing `/%[1]s \"Question\" \"Answer 1\" \"Answer 2\" \"Answer 3\"`\n" +
+		"Poll Settings provider further customisation, e.g. `/%[1]s \"Question\" \"Answer 1\" \"Answer 2\" \"Answer 3\" --progress --secret`. The available Poll Settings are:\n" +
+		"- `--secret`: Don't show who voted for what at the end\n" +
+		"- `--progress`: During the poll, show how many votes each answer option got\n"
+
+	// Parameter: Trigger
+	commandInputErrorFormat = "Invalid input. Try `/%[1]s \"Question\"` or `/%[1]s \"Question\" \"Answer 1\" \"Answer 2\" \"Answer 3\"`"
```
Hanzei: I din't know this working until now, but it allows us a cleaner code. See https://godoc.org/fmt#hdr-Printing

Hanzei:
@kaakaa This is ready for review

kaakaa:
Thanks for your works :+1:

While watching your code, an idea came up.

How about placing options immediately after /poll? E.g.: `/poll --progress --anonymous "Question" "Answer1" "Answer2"`

Since these options are used for determining the type of poll, it sounds good to place it immediately after the /poll command.
Also, when parsing the options, removing /poll first, and if starting with `--`, then parse the option, if other characters come, then we can pass the rest to the existing parsing logic .

I want to write code for confirming that this idea is consistent, but I'm sorry that I don't have time to do that until the weekend.

How do you think about this idea? If you're interested in this idea, I'll write code on this weekend.

Hanzei:
I don't think placing the options first makes parsing easier. And I think what most people comes to mind first is the question they want to ask. And only later they ask themself what Settings they want. So I would argue that its from a UX pov a better idea to place them last.

Which of my edge cases do you consider "wrong"?

kaakaa:
Sorry for my super late replying. Things have finally settled down at work.

I got it. With thinking about UX, I think that what you say is right.
On my experience, since many CLI tools place options following command name, I thought that it was normal to place optoins first. But I understand that my thought is developer's one.
So I agree with assumption that general users writes a question immediately after command. 

Let go with this. I'm sorry for taking time.
I will leave comments for some other parts.


FYI.
With paying attention only to the logic, I think that placing the options first makes easier to parse because the option **begins** with `-`.
I have written parsing code when writing the option first, and put it here.

refs: https://play.golang.org/p/aR60eWPFjRi

hanzei:
> Sorry for my super late replying. Things have finally settled down at work.

No problem. I'n not in a hurry.

I fixed the naming problem and added some code comments. Thank you for your code. Some day we might simplify the parsing.

This is ready for a new review.

PR #53, opened 2018-09-08
Respect Teammate Name Display

### LINKED ISSUE #52: Respect Teammate Name Display
The is a Mattermost setting called [Teammate Name Display](https://docs.mattermost.com/administration/config-settings.html#teammate-name-display) which "specifies how names are displayed in the user interface." We should respect this when displaying names.

Hanzei:
The best rule I came put with is:

- For author name: Use `GetDisplayName(model.SHOW_NICKNAME_FULLNAME)`
- For users at the end Poll screen:  Use `GetDisplayName(model.SHOW_USERNAME)`
As always this PR includes some refactoring. :wink: 

Fixes #52

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"errors"
+	"fmt"
 	"net/http"
 	"regexp"
 	"strings"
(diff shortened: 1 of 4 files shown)


server/plugin.go
```diff
@@ -53,12 +54,22 @@ func (p *MatterpollPlugin) ServeHTTP(c *plugin.Context, w http.ResponseWriter, r
...
+func (p *MatterpollPlugin) ConvertUserIDToDisplayName(userID string) (string, *model.AppError) {
 	user, err := p.API.GetUser(userID)
 	if err != nil {
 		return "", err
 	}
-	// NOTE: We should better fetch the server config and us this instead of model.SHOW_FULLNAME
-	nameFormat := model.SHOW_FULLNAME
-	return user.GetDisplayName(nameFormat), nil
+	displayName := user.GetDisplayName(*p.ServerConfig.TeamSettings.TeammateNameDisplay)
+	if displayName == user.Username {
```
kaakaa: When I tested in my local, I may have noticed useful spec of `TeammateNameDisplay` settings.
If server returns `@USERNAME`, client converts `@USERNAME` to the display_name according to `TeammateNameDisplay` settings in AccountSettings. Users can see display names in format users want to see.

So we have better to use `model.SHOW_USERNAME` and prepend `@` always. 

How do you think about this?

Hanzei: Oh, thank you for the hint. This kind of kills the purpose of this PR and make it do mostly refactoring. I updated my code and the description of this PR.

Hanzei:
This is ready for a second round of review.

server/plugin.go
kaakaa: Thank you for a lot of work and sorry for changing the purpose.
Updated code is LGTM, thanks !

PR #68, opened 2018-09-20
Refactoring

### LINKED ISSUE #55: Use mux for API routes
We should use https://github.com/gorilla/mux and ditch our self written router.

### LINKED ISSUE #60: Get rid of IDGenerator
With monkey patching for tests there is no need for a `IDGenerator`. To keep the code clean, we should remove it.

Hanzei:
This PR adds mux as a router for API requests, which makes the code much more clean and gets rid of regex for request parsing. But the means that we have to call `OnActivate()` every time we test the plugin. This is why I added a `setupTestPlugin()` function.

It also fixes two quite old problems with tests:
- There was a race condition in tests because some tests changed global variables. I added a `Copy()` method to `Poll` to get rid of this.
- We didn't compare `PostActionIntegrationResponse.Update` until now. This improves test by a lot.

Lately it makes use of https://github.com/mattermost/mattermost-server/pull/9351

Fixes #55 
Fixes #60


Hanzei:
12 hours later isn't late. Your reviews are great and I'm happy to wait a couple of days. No hurry.

Hanzei:
I will also include  #60 into this PR.

Hanzei:
This is ready for review

server/api_test.go
```diff
@@ -77,66 +74,63 @@ func TestServeFile(t *testing.T) {
...
 	w := httptest.NewRecorder()
-	r := httptest.NewRequest("POST", "/logo_dark.png", nil)
+	r := httptest.NewRequest("POST", fmt.Sprintf("/%s", iconFilename), nil)
 	p.ServeHTTP(nil, w, r)
 
 	result := w.Result()
 
 	bodyBytes, err := ioutil.ReadAll(result.Body)
-	assert.Nil(err)
+	require.Nil(t, err)
```
kaakaa: Do you have any policy on which of `assert` and `require` should be used?
I may understand the difference between `assert` and `require`, but I don't know which I should use when asserting.

Hanzei: I use `require` when it doesn't make sense to continue a test if a check fails. This is mostly the case when a struct is returned and I want to check its fields. Take https://github.com/matterpoll/matterpoll/blob/55_mux/server/command_test.go#L185 as am example. If `r` is nil the rest of the test will just throw a NPE. Its better to stop the run there and exit. This makes debugging tests easier.

kaakaa: Thank you for explaining it. I got it and feel making sense 👍

PR #70, opened 2018-09-20
Fix trigger word in README

kaakaa:


kaakaa:
> What do you think about adding a section called Settings?

That is good idea! I added `Settings` section. PTAL.

PR #77, opened 2018-09-26
Allow System Admins to delete and end polls

### LINKED ISSUE #74: Permission for Admins to End or Delete Polls
It would be nice not only for the creator of a poll but also for the admins to be able to end or to delete a poll.

Hanzei:
@Twilek-de feel free to test this PR

Fixes #74

### DIFF
--- a/server/api_test.go
+++ b/server/api_test.go
@@ -267,6 +267,21 @@ func TestHandleEndPoll(t *testing.T) {
 	expectedPost1 := &model.Post{}
 	model.ParseSlackAttachment(expectedPost1, expectedattachments1)
 
+	api9 := &plugintest.API{}
+	api9.On("KVGet", samplePollID).Return(samplePollWithVotes.Encode(), nil)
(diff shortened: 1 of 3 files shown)


server/api_test.go
```diff
@@ -267,6 +267,18 @@ func TestHandleEndPoll(t *testing.T) {
 	expectedPost1 := &model.Post{}
 	model.ParseSlackAttachment(expectedPost1, expectedattachments1)
 
+	api9 := &plugintest.API{}
```
Hanzei: My policy for naming and sorting sub tests and apis:
- Keep the tests and apis in the same orders as they are used in the code
- Bump api numbers for new tests

2/5 if this is a good policy

kaakaa: To be honest, I got a bit confusing for mixed numbering of api veriables.
And since api numbers is mixed, when adding apis, next number will not be immediately understood.

But I don't have nicer idea, so I think that your policy is better realistic solution for now. 👍

PR #81, opened 2018-10-04
Sync read and write to config

hanzei:
Since plugin methods are called asynchrony `Config` could be changed any time. This PR fixes this by using a lock. It is pretty much a downstream of https://github.com/mattermost/mattermost-plugin-demo/pull/7.

The second commit fixes come linter issues. Some day I will learn to submit separate PRs for this. Sorry @kaakaa.

### DIFF
--- a/server/configuration.go
+++ b/server/configuration.go
@@ -1,31 +1,70 @@
 package main
 
-import "errors"
+import (
+	"github.com/pkg/errors"
+)
 
+// Config captures the plugin's external configuration as exposed in the Mattermost server
+// configuration, as well as values computed from the configuration. Any public fields will be
+// deserialized from the Mattermost server configuration in OnConfigurationChange.
 type Config struct {
 	Trigger string
 }
 
+// OnConfigurationChange loads the plugin configuration, validates it and saves it.
 func (p *MatterpollPlugin) OnConfigurationChange() error {
-	c := &Config{}
-	if err := p.API.LoadPluginConfiguration(c); err != nil {
-		return err
+	config := new(Config)
+
+	if err := p.API.LoadPluginConfiguration(config); err != nil {
+		return errors.Wrap(err, "failed to load plugin configuration")
 	}
 
-	if c.Trigger == "" {
+	if config.Trigger == "" {
 		return errors.New("Empty trigger not allowed")
 	}
 
 	if p.Config != nil {
 		if err := p.API.UnregisterCommand("", p.Config.Trigger); err != nil {
-			return err
+			return errors.Wrap(err, "failed to unregister old command")
 		}
 	}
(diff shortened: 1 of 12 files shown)


server/configuration.go
```diff
@@ -1,31 +1,70 @@
...
+	p.setConfiguration(config)
 	return nil
 }
+
+// getConfiguration retrieves the active configuration under lock, making it safe to use
+// concurrently. The active configuration may change underneath the client of this method, but
+// the struct returned by this API call is considered immutable.
+func (p *MatterpollPlugin) getConfiguration() *Config {
+	p.configurationLock.RLock()
+	defer p.configurationLock.RUnlock()
```
kaakaa: Sorry, I don't know too much about mutex, so I don't understand what does this lock improve.

If calling RUnlock here, RLock is immediately unlocked in the end of this method.
So I think that we can update config value at any time regardless of this changes.

```
config := p.getConfiguration()
config.Trigger = "dummy"
```

Is this PR intended to prevent config values from being changed during evaluating simple if statement and return statement?

hanzei: > If calling RUnlock here, RLock is immediately unlocked in the end of this method.
So I think that we can update config value at any time regardless of this changes.

That is is correct. A client of this method can change the configuration value at any time. But besides of a separate package, the is no way to avoid this. The comment of this method explains this quite well
> getConfiguration retrieves the active configuration under lock, making it safe to use
concurrently. The active configuration may change underneath the client of this method, but
the struct returned by this API call is considered immutable.

> Is this PR intended to prevent config values from being changed during evaluating simple if statement and return statement?

Correct. This PR ensured that a configuration doesn't get changed while a methods evaluates it. The biggest problem is in `ExecuteCommand()`. We evaluate the configuration values multiple times, but a change can happen every moment.

hanzei:
My first code was rather sloppy and missed out on some configuration uses. It fixes those and changed `Config`to `configuration` to stay consistent with the upstream repo. Please review again,

server/configuration.go
kaakaa: It all makes sense now.
Thank you for the clarification!

server/configuration.go
```diff
@@ -1,31 +1,70 @@
...
 
-	if c.Trigger == "" {
+	if configuration.Trigger == "" {
 		return errors.New("Empty trigger not allowed")
 	}
 
-	if p.Config != nil {
-		if err := p.API.UnregisterCommand("", p.Config.Trigger); err != nil {
-			return err
+	if p.getConfiguration().Trigger != "" {
```
kaakaa: Should we avoid to call `getConfiguration` twice here?
Or `OnConfigurationChange` isn't called concurrently?

PR #85, opened 2018-10-08
Add store

### LINKED ISSUE #59: Add unique prefix to DB keys
We store polls into the DB by using `pollID` as a key. This limits our DB to only containing polls. By adding a prefix e.g. `pollID_` to the key, we avoid this problem.

This would of cause break all previous polls. But we are still early in development and I don't consider this a blocking problem.

This should be done before v1.0

### LINKED ISSUE #62: Save Matterpoll version for migration
We should store the Matterpoll version in the DB. This allows us to do some kind of migration in future versions.

hanzei:
This PR aims to move all DB operations into the methods of a `Store` struct. This keeps the code DRY and allows for easier changes like adding a prefix to a key.

The PR also introduces a `SystemStore` which is capable of storing the plugin version.

Lastly this PR handles the migration from an old schema to a new schema.

### Further thoughts 

The current state is more of a proposal. Specially the test site is not finished. Further ideas are:
- Convert `Store` into an interface
- Add a `APIStore` struct
- Add a `TestStore` struct

I would like to bring this proposal up for discussion. All feedback is welcome!

### Todo
- [x] Cleanup tests in `plugin` package
- [x] Design mock store (See "Further thought")
- [x] Enable poll store with `pollPrefix`

**Important note: This PR introduces breaking changes. Polls created prior to this PR will not work any longer!**

Fixes #62 
Fixes #59


hanzei:
@kaakaa Would you give this one a first round of review. I would like to hear your thoughts about this. I know, its a little over specd, but also powerful. With this framework we can easily do data migration.

server/plugin/api.go
```diff
@@ -90,22 +83,20 @@ func (p *MatterpollPlugin) handleVote(w http.ResponseWriter, r *http.Request) {
 	}
 
 	hasVoted := poll.HasVoted(userID)
-	err := poll.UpdateVote(userID, optionNumber)
-	if err != nil {
+	if err := poll.UpdateVote(userID, optionNumber); err != nil {
```
kaakaa: Should we avoid idiomic error handling?  `if err := method(); err != nil { ... }`

server/store/apistore/poll_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/matterpoll/matterpoll/server/utils/testutils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPollStoreGet(t *testing.T) {
+	t.Run("all fine", func(t *testing.T) {
+		api := &plugintest.API{}
+		api.On("KVGet", testutils.GetPollID()).Return(testutils.GetPoll().EncodeToByte(), nil)
+		store := setupTestStore(api)
```
kaakaa: Should call `defer api.AssertExpectations(t)` in all test cases.

server/store/apistore/poll.go
```diff
@@ -0,0 +1,43 @@
+package apistore
```
kaakaa: What does `api` of `apistore` represent?
I think that `pluginstore` or `kvstore` is better than `apistore`.

server/store/apistore/poll.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (s *PollStore) Save(poll *poll.Poll) error {
+	// err := s.api.KVSet(pollPrefix+poll.ID, poll.Encode())
+	if err := s.api.KVSet(poll.ID, poll.EncodeToByte()); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (s *PollStore) Delete(poll *poll.Poll) error {
```
kaakaa: I like that the argument of this func is `id string`.

server/store/apistore/poll_test.go
hanzei: Good catch! :+1:

server/store/apistore/poll.go
hanzei: I agree and changed it to `kvstore`.

server/plugin/api.go
hanzei: I'm fine using `if err := method(); err != nil { ... }`. You might also want to take a look at https://developers.mattermost.com/blog/idiomatic-error-handling.

server/store/apistore/poll.go
hanzei: I'm not sure I can follow you.

hanzei:
@kaakaa This is ready for a final round of review.

server/plugin/plugin_test.go
```diff
@@ -93,12 +105,31 @@ func TestPluginOnActivate(t *testing.T) {
 			}
 		})
 	}
+	t.Run("NewStore() fails", func(t *testing.T) {
+		api := &plugintest.API{}
+		api.On("GetServerVersion").Return(minimumServerVersion)
+		defer api.AssertExpectations(t)
+
+		patch := monkey.Patch(kvstore.NewStore, func(plugin.API) (store.Store, error) {
```
hanzei: I know this is ugly, but I didn't find another way to properly test this

server/plugin/versions.go
```diff
@@ -7,11 +7,4 @@ var apiVersions = []string{
...
 // CurrentAPIVersion is the newest API version
-var CurrentAPIVersion = apiVersions[0]
-
-var dataSchemaVersions = []string{
-	"v1",
-}
-
-// CurrentDataSchemaVersion is the newest data schema version
-var CurrentDataSchemaVersion = dataSchemaVersions[0]
+var CurrentApiVersion = apiVersions[0]
```
hanzei: I tent to remove this file completely. I hope we will never have more then two api version running and don't need a list for them.

server/poll/transform.go
```diff
@@ -64,7 +64,7 @@ func (p *Poll) makeAdditionalText(numberOfVotes int) string {
 
 	lines := []string{"---"}
 	if len(settingsText) > 0 {
-		lines = append(lines, fmt.Sprintf("**Poll settings**: %s", strings.Join(settingsText, ", ")))
+		lines = append(lines, fmt.Sprintf("**Poll Settings**: %s", strings.Join(settingsText, ", ")))
```
hanzei: I changed this to be in line with the Readme. Hope you are fine with this @kaakaa.

server/plugin/api.go
kaakaa: I got it. You're right. There is no need to fix this.

server/plugin/versions.go
kaakaa: I agree with you.

server/store/kvstore/upgrade.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+func (s *Store) UpdateDatabase() error {
+	v, err := s.System().GetVersion()
+	if err != nil {
+		return err
+	}
+	// If no version is set, set to to the newest version
+	if v == "" {
+		// TODO: Dont hardcode this
```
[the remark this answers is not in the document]
kaakaa: `version.go`? Are you saying about `manifest.go`?
If so, I think that `manifest.go` should be in `plugin` package because of its meaning...

Would you like to add the argument `pluginVersion string` to `UpdateDatabase` function? That argument has the meaning of what version the store migrates to.
The value of `plugingVersion` is given in `OnActivation` func and via `NewStore` func.

How do you think about this?

If you don't like this idea, I'm fine that we move `manifest.go` to other package. However, in that case we don't need to change `poll` package name.

server/plugin/plugin_test.go
kaakaa: Would you like to add `UpdateDatabase` func to `store.Store` interface? If doing so, we can modify `NewStore` func to return error by mocking `UpdateDatabase`.
I do not feel strange even if that `UpdateDatabase` func is in the `store.Store` interface.

How does it sounds? If you agree my idea, see also [this comments](https://github.com/matterpoll/matterpoll/pull/85#discussion_r228826995).

server/store/kvstore/upgrade.go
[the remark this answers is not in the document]
hanzei: :+1: on adding `pluginVersion string` as an argument to `UpdateDatabase`.

server/plugin/plugin_test.go
hanzei: Sure we can add `UpdateDatabase` to `store.Store`. But how does this helps us for this problem? `kvstore.NewStore` will always return a `kvstore.Store`.
I don't think I get what you are up to.

kaakaa: Ah..sorry. I misunderstood that we can mock `UpdateDatabase` if adding it to interface...
But we can't do that. you are right.

Now I don't have a better solutions. So I'm fine with present code.

hanzei:
I added `PluginVersion` to `UpgradeDatabase()`. @kaakaa can you please review again?

PR #97, opened 2018-10-30
Fix release file name for deploy

kaakaa:


hanzei:
@kaakaa Mind if we release v1.0.1 with this? This way we can text it again and if people really want a CI build release, they can use v1.0.1.

hanzei:
Mind if I delete https://github.com/matterpoll/matterpoll/releases/tag/deploy-test?

kaakaa:
Yes, we should release new patch version (v1.0.1). because we better leave your uploaded binary.

hanzei:
See https://github.com/matterpoll/matterpoll/pull/98

PR #104, opened 2018-11-04
use plugins folder

### LINKED ISSUE #91: logo_dark.png not found in the URL
The image cannot be found in the current URL, leading to the profile picture not showing properly.

zetaab:
fixes #91

### DIFF
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -15,7 +15,7 @@ const (
 	infoMessage = "Thanks for using Matterpoll v" + PluginVersion + "\n"
 
 	iconFilename = "logo_dark.png"
-	iconPath     = "plugins/" + PluginId + "/"
+	iconPath     = "/" + PluginId + "/"
 
 	voteCounted = "Your vote has been counted."
 	voteUpdated = "Your vote has been updated."
@@ -52,7 +52,7 @@ func (p *MatterpollPlugin) handleInfo(w http.ResponseWriter, r *http.Request) {
 
 func (p *MatterpollPlugin) handleLogo(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Cache-Control", "public, max-age=604800")
-	http.ServeFile(w, r, iconPath+iconFilename)
+	http.ServeFile(w, r, *p.ServerConfig.PluginSettings.Directory+iconPath+iconFilename)
 }
 
 func (p *MatterpollPlugin) handleVote(w http.ResponseWriter, r *http.Request) {


zetaab:
this is deployed to our test and production mattermost. Works in both now

hanzei:
@zetaab Thanks for the PR.

While this fixed your problem with a change plugin directory it does not solve the issue with different working directories. Your code still assumes that the plugins gets executed within `/mattermost/` or `/mattermost/bin/`. If this root cause gets fixed, this will also fix your problem.

Because of this I tend to favour a PR which fixes the root cause instead of a symptom.

zetaab:
@hanzei mattermost server binary is using this same method to locate plugins folder. I do not see any reason why we should something else that is used in mattermost server itself.

hanzei:
> mattermost server binary is using this same method to locate plugins folder.

I don't get that your are saying. Can you point me to the code?

zetaab:
```
 matterpoll [fixpath●●] % oc rsh mattermost-6bf564f6f5-97pm4
$ ps x
   PID TTY      STAT   TIME COMMAND
     1 ?        Ss     0:00 /bin/bash /docker-entry.sh
     9 ?        Sl     0:05 ./mattermost server
    40 ?        Sl     0:00 /tmp/com.github.matterpoll.matterpoll/server/dist/plugin-linux-amd64
    52 ?        Ss     0:00 /bin/sh -c TERM="xterm-256color" /bin/sh
    58 ?        S+     0:00 /bin/sh
    73 ?        Ss     0:00 /bin/sh -c TERM="xterm-256color" /bin/sh
    79 ?        S      0:00 /bin/sh
    80 ?        R+     0:00 ps x
```

mattermost binary is located in `/mattermost/bin/` folder. Plugins are now located under `/tmp` 

mattermost config file contains:
```
    "PluginSettings": {
        "Directory": "/tmp/",
...
```

What else this should support. I do not understand? Now you can locate your plugin directory to anywhere in running server.

kaakaa:
I think thit PR is making sense. mattermost-server seems to locates plugin assets to directory specified `PluginSettings.Directory`.

https://github.com/mattermost/mattermost-server/blob/master/app/plugin_install.go#L76
```go
pluginPath := filepath.Join(*a.Config().PluginSettings.Directory, manifest.Id)
err = utils.CopyDir(tmpPluginDir, pluginPath)
```

When plugins look for their assets from `PluginSettings.Directory` value, I think that it could be found even if the path specified in `PluginSettings.Directory` is an absolute path or a relative path.

If there is an example that my assumption is broken, I would like to know that.

zetaab:
@kaakaa yes, it supports both relative and absolute paths. Like I commented I used `/tmp` and currently in our production environment I am using `../clients/plugins`. Both working well

hanzei:
There a (at least) two circumstances which lead to #91:
1. `PluginSettings.Directory` is not `./plugins`
2. The working directory of Mattermost is not the "root" Mattermost directory.
There might be more way that lead to this problem.
Your PR fixes 1.But the root problem is that Matterpoll tries to serve files relative to _its working directory_. This is inherently bad and should be changes. By serving files relative to _the path of the executable_  we can solve the root cause.

I'm working on a PR to change this. If we merge your PR, your changes will just be overridden soon.

kaakaa:
@hanzei  Hmm...sorry, I've not sure what is problem yet. I think that zetaab's PR can fix No.2 circumstance.

Suppose we are on following settings.

* "root" Mattermost directory is `/opt/mattermost`
* Working directory of Mattermost is `/usr/local/mm`
  * Before runnig mattermost, we must copy assets of mattermost client in `/usr/local/mm/client` and make directory `/usr/local/mm/dummy/plugins` and `/usr/local/mm/client/dummy/plugins` 
* `PluginSettings.Directory` = `./dummy/plugins`
* Run mattermost by executing `/opt/mattermost/bin/mattermost`

When uploading a plugin to Mattermost, mattermost-server places the plugin in `/usr/local/mm/dummy/plugins/${PLUGIN_ID}`.  Then, if plugin try to read a file from PluginSettings.Direcotory, plugin can find it from `/usr/local/mm/dummy/plugins/${PLUGIN_ID}`.

With this settings, this PR can fix this issue. I confirmed this by changing `WorkingDirectory` in `mattermost.service` in [official setup document](https://docs.mattermost.com/install/install-ubuntu-1804.html).

What settings can I reproduce the problem you said?

----

Of course, your PR looks fine (zetaab's PR looks fine too). The difference between hanzei's PR and zetaab's PR is whether the file searching way is bottom-up or top-down. The both is same for me from the viewpoint of how to find icon file.

PR #112, opened 2018-11-15
Display "0 vote" instead of "0 votes".

Suliv4n:


hanzei:
Thanks for the PR @Suliv4n!
I'm not an native speaker, but I think "0 votes" is correct. See https://english.stackexchange.com/questions/38293/why-is-zero-followed-by-a-plural-noun.

Suliv4n:
Oh, ok i didn't know that. Thanks for the information.

PR #122, opened 2018-12-14
Fix log messages

hanzei:


hanzei:
Thanks for the review @kaakaa! Please give this another round.

PR #124, opened 2019-01-05
Update dependencies

hanzei:
This will require a dev environment with the latest `master` of mattermost-server to run `make deploy` because of https://github.com/mattermost/mattermost-plugin-sample/pull/23.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 GO ?= $(shell command -v go 2> /dev/null)
 DEP ?= $(shell command -v dep 2> /dev/null)
 NPM ?= $(shell command -v npm 2> /dev/null)
-HTTP ?= $(shell command -v http 2> /dev/null)
 CURL ?= $(shell command -v curl 2> /dev/null)
 MANIFEST_FILE ?= plugin.json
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -163,7 +153,11 @@ ifneq ($(HAS_WEBAPP),)
...
 
-# generate store mocks
+## Generate store mocks
 .PHONY: store-mocks
 store-mocks:
 	mockery -name=".Store" -dir server/store -output server/store/mockstore/mocks -note 'Regenerate this file using `make store-mocks`.'
+
+# Help documentatin à la https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
+help:
+	@cat Makefile | grep -v '\.PHONY' |  grep -v '\help:' | grep -B1 -E '^[a-zA-Z_.-]+:.*' | sed -e "s/:.*//" | sed -e "s/^## //" |  grep -v '\-\-' | tac | awk 'NR%2{printf "\033[36m%-30s\033[0m",$$0;next;}1' | sort
```
kaakaa: `tac` is not default command for macOS, so `make help` may fail in macOS.  https://unix.stackexchange.com/a/114042
I think that `perl -e 'print reverse <>' ` is more general. (But `perl` is not POSIX, so my suggestion is not perfect)

What do you think about supports of macOS?

hanzei: I would like to stay consistent with https://github.com/mattermost/mattermost-plugin-sample.

0/5 about the out of the box support of macOS. But also 0/5 on the POSIX compatibility. What do you think?

kaakaa: Got it. Now, I'll merge this PR, because `make help` is not critical for development.

---

But I hope that plugin dev tools has compatibility with macOS, because I mainly use macOS for development of mattermost plugin.

In [this comments](https://github.com/mattermost/mattermost-plugin-sample/pull/19#issuecomment-436811575), lieut-data  said that he can replace `tac` with `awk`, and `awk` is POSIX command. I will ask him replacing `tac` with `awk`.

PR #133, opened 2019-02-28
Localizations

### LINKED ISSUE #119: translations / localizations
I'd love to see this plugin in non-English languages. Happy to provide a German translation but unsure how and if it's possible at all right now. The best way might be if the plugin uses mattermost's language setting.

hanzei:
This PR add support for localization.

Feedback on this PR from any perspective (user, dev, translator) is highly appreciated.

If you are a plugin author and looking for some boilerplate code, take a look at https://github.com/matterpoll/matterpoll/pull/133/commits/54617ce30edf17b6c365271eb108e9a2f808e12d.

## General translation rules
- When a translation is missing, use the English one.
- For texts displayed to only one user e.g. ephemeral messages, use the users local.
- For texts displayed to multiple users e.g. use [the Default Server Language](https://docs.mattermost.com/administration/config-settings.html#default-server-language). (I'm not 100% happy with that. I would love to hear other ideas on this)

There is currently only a German translation. People testing this PR must set there localization to German.

I would love feedback from speakers of other languages. Does this translation strings work for you? An additional challenge are CJK languages (:wink: @kaakaa)

## Technical notes

This is pretty much a first try on localization of plugins. I tried to "get something" done.

I use the same library (https://github.com/nicksnyder/go-i18n) as the MM server, but version 2. The main difference between V1 and V2 are:
- All default strings are kept inside the codebase. This improves the dev experience greatly. No more tapping to `en.json`.
- Improved support for plural forms
- Removed the global state
- There are now **two** translation file for every language:
 - `active.de.json`: This contain all translated messages
 - `translate.de.json`: This contains to messages, that have to be translated. This file get auto populated every time a messages in the source changes. (This is awesome). New messages also get auto populated into this file, so you easy to track what needs to be translated.
- New CLI tools to support the two files (They require some time go get used to, I admit)

I did keep the messages in global scope. This allows me to use them in tests. This might not be suitable for everybody.

I opted to use `json` instead of the default `toml`. 0/5 if this is a good idea. Might be worth to go with the standard.

## How to do translation?

This is a bit complicated, but:
- We can add commands to the make file
- Running the command inside `i18n` makes the easier. But I wanted to be able to run them from the root directory.
- Using `toml` instead of `json` would make this easier.

### Translating a new language

Replace `de` with your local!

`touch asserts/i18n/translate.de.json`

`goi18n merge -format json -outdir assets/i18n/ assets/i18n/active.en.json
assets/i18n/translate.de.json`

After `translate.de.json` has been translated, rename it to `active.de.json`.

See also https://github.com/nicksnyder/go-i18n#translating-a-new-language

### Translating new messages

`goi18n extract -format json -outdir assets/i18n/ server/`

`goi18n merge -format json -outdir assets/i18n/ i18n/active.*.json`

Translate your messages

`goi18n merge -format json -outdir assets/i18n/ i18n/active.*.json i18n/translate.*.json`

See also https://github.com/nicksnyder/go-i18n#translating-new-messages

## Open problems
The following things are not translatable, due to technical limitations of Mattermost:
- System console settings page
- Slash command help text

## ToDo
- [x] Load all translation files inside `i18n`
- [x] Fix tests

Fixes #119


server/plugin/command.go
```diff
@@ -34,14 +72,39 @@ func (p *MatterpollPlugin) ExecuteCommand(c *plugin.Context, args *model.Command
...
+			DefaultMessage: commandHelpTextOptions,
+			TemplateData:   map[string]interface{}{"Trigger": configuration.Trigger},
+		}) + "\n"
+		msg += localizerUser.MustLocalize(&i18n.LocalizeConfig{
+			DefaultMessage: commandHelpTextPollSettingIntroduction,
+			TemplateData:   map[string]interface{}{"Trigger": configuration.Trigger},
+		}) + "\n"
+		msg += "- `--anonymous`: " + localizerUser.MustLocalize(&i18n.LocalizeConfig{DefaultMessage: commandHelpTextPollSettingAnonymous}) + "\n"
+		msg += "- `--progress`: " + localizerUser.MustLocalize(&i18n.LocalizeConfig{DefaultMessage: commandHelpTextPollSettingProgress}) + "\n"
+		msg += "- `--public-add-option`: " + localizerUser.MustLocalize(&i18n.LocalizeConfig{DefaultMessage: commandHelpTextPollSettingPublicAddOption})
```
hanzei: Any ideas on how to make this prettier?

server/plugin/api.go
```diff
@@ -88,41 +122,51 @@ func (p *MatterpollPlugin) handleVote(w http.ResponseWriter, r *http.Request) {
...
+		response.EphemeralText = commandErrorGeneric.Other
+		writePostActionIntegrationResponse(w, response)
+		return
+	}
+	localizer := i18n.NewLocalizer(p.bundle, user.Locale)
+
 	poll, err := p.Store.Poll().Get(pollID)
 	if err != nil {
-		response.EphemeralText = commandGenericError
+		response.EphemeralText = localizer.MustLocalize(&i18n.LocalizeConfig{DefaultMessage: commandErrorGeneric})
```
hanzei: Might be worth to move ` localizer.MustLocalize(&i18n.LocalizeConfig{}` into a separate method.

scottleedavis: I think so too.   Would also shorten up the above prettier issue.

server/utils/file.go
```diff
@@ -0,0 +1,15 @@
+package utils
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// GetPluginRootPath ToDo
+func GetPluginRootPath() string {
```
scottleedavis: Since there is only one func in this file, perhaps it is better to put in i18n.go? 3/5

server/plugin/command.go
kaakaa: I'm fine with messages to be translated per line. 2/5

i.e.:
```
msg += localizerUser.MustLocalize(&i18n.LocalizeConfig{
    DefaultMessage: commandHelpTextPollSettingAnonymous,
    TemplateData:   map[string]interface{}{"AnonymousSetting": "--anonymous"},
})
msg += localizerUser.MustLocalize(&i18n.LocalizeConfig{
    DefaultMessage: commandHelpTextPollSettingProgress,
    TemplateData:   map[string]interface{}{"ProgressSetting": "--pregress"},
})
msg += localizerUser.MustLocalize(&i18n.LocalizeConfig{
    DefaultMessage: commandHelpTextPollSettingPublicAddOption,
    TemplateData:   map[string]interface{}{"PublicAddOptionSetting": "--public-add-option"},
})
```

```
commandHelpTextPollSettingAnonymous = &i18n.Message{
    ID:    "command.help.text.pollSetting.anonymous",
	Other: "- {{.AnonymousSetting}}: Don't show who voted for what\n",
}
...
```

server/utils/file.go
kaakaa: This func is used for getting logo image, so moving this to `i18n.go` makes little confusing.
https://github.com/matterpoll/matterpoll/pull/133/files#diff-4444f0caa92e553b5c018c908e90e5e6R107

I think that it had better to leave this here for now.

Makefile
```diff
@@ -93,6 +93,7 @@ bundle:
 	mkdir -p dist/$(PLUGIN_ID)
 	cp $(MANIFEST_FILE) dist/$(PLUGIN_ID)/
 	cp assets/logo_dark.png dist/$(PLUGIN_ID)/
+	cp -r i18n/ dist/$(PLUGIN_ID)/
```
kaakaa:
```suggestion
	cp -r i18n/ dist/$(PLUGIN_ID)/i18n
```

Original code doesn't create `i18n` directory, so starting plugin is failed.

server/plugin/plugin.go
```diff
@@ -27,6 +28,8 @@ type MatterpollPlugin struct {
 	// setConfiguration for usage.
 	configuration *configuration
 	ServerConfig  *model.Config
+
+	bundle *i18n.Bundle
```
kaakaa: I want to reduce places where `i18n.NewLocalizer` func is called, because many localizer with different locale will make me confusing.
So we should have new wrapper i18n.Bundle. 1/5

e.g.:
```
type MatterPollLocalizer struct {
    bundle *i18n.Bundle
    localizer *i18n.Localizer  // Localizer made by DefaultServerLocale
}
func (l *MatterPollLocalizer) localize(...) { ... }
func (l *MatterPollLocalizer) localizeWithUserLocale(u *model.User, ...) { ... }
```

This comment might be related to other comments https://github.com/matterpoll/matterpoll/pull/133/files#r261197956.

kaakaa:
Great!
I translated some messages to Japanese, run it, and It seemed to work fine 👍 

About `json` or `toml`, if we want to add any comment lines in `active.*.[json|toml]`, we must use `toml`. But for now, we don't need to add comments, so using `json` would have no problem.

server/utils/file.go
scottleedavis: Oh!  I naively assumed it was created for this single purpose.

server/plugin/plugin.go
hanzei: With the current code, the lifetime of a `i18n.Localizer` is one hook (e.g. `serverHTTP`)

If I understand you correctly, your proposal is to have it the lifetime of the whole plugin instead. Could you elaborate a bit, why do you think this is needed? If I understand the idea of `go-i18n` correctly, `i18n.Localizer` are supposed to be ephemeral.

One additional thought: There are only two places where not the user local is used: When a poll is created and when its deleted. I would argue, that not so often.

To get this straight: I'm not against a Wrapper around `i18n.Localizer`.

server/plugin/api.go
hanzei: Let's resolve the other conversation first and see if this problem still exists.

server/plugin/command.go
hanzei: What is the great difference between your proposal and what I did besides moving the setting name inside the translation?

server/plugin/plugin.go
kaakaa: Thank you for your confirmation.
I did't care about the lifetime of i18n.Localizer and the idea of `go-i18n`. (but I supposed that having the lifetime of whole plugin has no problem.)

I cared about only readability. With one-hook (current approach), sometimes we would check which locale the variable has. 

```
localizer := i18n.NewLocalizer(p.bundle, locale)

...(many sentences)...

localizer.MustLocalizer(message_id, params)  // ...which locale does "localizer" has?
```

With my suggestion, we would be able to know easily which locale will be used.
```
p.locaizer.localize(message_id, params)
p.locaizer.localizeWithUserLocale(user, message_id, params)
```

I don't have strong opinion and I don't intend to reject current approach, because I haven't use `go-i18n` library. 


> One additional thought: There are only two places where not the user local is used: When a poll is created and when its deleted. I would argue, that not so often.

Two places? When searching this branch with `i18n.NewLocalizer`, I got 12 places using it.

hanzei:
@jasonblais 
 
> 1 When a string gets updated, it's added to translate.de.json. Does it also get removed from active.de.json?

Yes, it get's removed from `active.de.json`.

> 2 Is this due to a technical limitation?

With server only plugins a post will be shown in the same way to every user. This is something we might want to change in the future, but is out of scope for now IMO.

Hence the translation must be the same for everyone. This brings up the interesting question on what should be the translation to show. I see three possibilities:
1. The users local
2. The Default Server Language
3. Add a plugin setting to change this.

I opted to with 2. because 1. doesn't makes sense if you are e.g. the only german guys on the team. And 3. might be tedious for admins to set for every plugin. 1/5

Site note: One other area where the same text is shown to every use are slash command help texts.

> 3 - Is the proposal to submit translations to individual plugins, and not manage them via translate.mattermost.com?

No. Since Elisa asked the same question and it's a bit out of scope for this PR, I answered it on https://community.mattermost.com/core/pl/na77w4z963fgjyf6zcsm6w3ppo.

> 4 - Translating messages involved some Go commands. Is this optional, or can they be directly translated by non-technical users as well?

Translating messages only requires the `goi18n` command which is a go tool. No further Go commands are needed. Therese command might be abstracted via the Makefile.

jasonblais:
1 - Excellent
2 - Okay, we can improve on it later. For now, if the message has to be the same for all users, agree with option 2.
3 - Makes sense. I'm not sure adding them to translate.mattermost.com is the best approach, will need to think about it more. So this is great for V1.

So 1 -> 3 are fine.

4 - My worry is that non-technical users who could help with translations won't know how to run the command.

hanzei:
> 4 - My worry is that non-technical users who could help with translations won't know how to run the command.

Could you elaborate a bit? What skills does a "non-technical user" have? Do they know what a shell is? Can they run a `make` command?

jasonblais:
> Could you elaborate a bit? What skills does a "non-technical user" have? Do they know what a shell is? Can they run a make command?

Sure, that's a good question. We can have (and likely have) translators who have never even used a terminal.

That said, our main audience is in DevOps so we can start with what have now, especially since it's based on https://github.com/nicksnyder/go-i18n. We can look at making the experience easier if there's a need for it later.

hanzei:
@jasonblais I would like to hear you thoughts about 2. If you ignore all technical limitations, what messages should be translated with what language? What would be your "algorithm" to decide this?

server/plugin/plugin.go
hanzei: What if I just rename `localizer` to `localizerUser`? E.g.:

```
localizerUser := i18n.NewLocalizer(p.bundle, locale)

...(many sentences)...

localizerUser.MustLocalizer(message_id, params)  // ...which locale does "localizer" has?
```
Would that help in terms of readability?
 
> Two places? When searching this branch with `i18n.NewLocalizer`, I got 12 places using it.

Sorry, my answer was ambiguous. The point is: A localizer is using the users local every time. But in two places it uses the system language instead.

i18n/active.de.json
```diff
@@ -0,0 +1,123 @@
+{
```
hanzei: I've been thinking about this a bit more and propose to use either `toml` or `yaml` instead of `json` because of two reasons:
- I want comments in my files
- In `json` you have to escape `"` and this sucks.

I'm a bit leaning towards using `toml` because it's the default for `go-1i8n`, but really I'm 0/5.

@kaakaa @jasonblais @scottleedavis Thoughts?

jasonblais:
For 2, the principles for me are:

1. Anything in the user interface, e.g. end user UI, admin console, .. should be localized to the user's locale in Account Settings > Language > Display
2. Anything used for troubleshooting/debugging, e.g. server logs, database logs, etc should not be translated, because it makes web search useless.

i18n/active.de.json
jasonblais: My thought: We'd need to make a decision on whether these go to the Translations server or not. If not, I'm 0/5 on the chosen formatting.

scottleedavis: 0/5

server/plugin/plugin.go
kaakaa: Renaming improve somewhat readability. But if just changing variable name, we'll must remenber the convention of naming when creating new localizer variables. So I like switching locale by method name.
My suggestion would be non-blocking comments, and your code is fine for first commit of localization. So if you don't think that my suggestion is good, please go ahead to merge. Of course, I welcome to continue discussing about this topic.

> Sorry, my answer was ambiguous. The point is: A localizer is using the users local every time. But in two places it uses the system language instead.

It seems that Localizer with server locale is used for `handleVote` and `handleAddOption` too? It will be removed?

i18n/active.de.json
kaakaa: I'm also on 0/5 too.

> I want comments in my files

What kind of comments do you want to add? My decision point would be here.

> In json you have to escape " and this sucks.

I strongly agree with you. But `yaml` and `toml` seem to have same problem...

hanzei: > I strongly agree with you. But yaml and toml seem to have same problem...

It think `yaml` doesn't uses " for marking strings. See https://gist.github.com/oconnor663/9aeb4ed56394cb013a20

hanzei: Head up: I'm thinking about this and will post a proposal, when I'm done thinking :wink:

hanzei:
I did a lot a refactoring. @kaakaa Would you please take a second look.

I tried fixing all tests, but there are a few that still fail with a suspicious message, e.g.:
```
    --- FAIL: TestPollToPostActions/Multipile_questions,_settings:_anonymous,_public-add-option (0.01s)
        transform_test.go:243: 
            	Error Trace:	transform_test.go:243
            	Error:      	Not equal: 
            	            	expected: []*model.SlackAttachment{(*model.SlackAttachment)(0xc0000d7e60)}
            	            	actual  : []*model.SlackAttachment{(*model.SlackAttachment)(0xc000229b00)}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -11,3 +11,3 @@
            	            	   TitleLink: (string) "",
            	            	-  Text: (string) (len=70) "---\n**Poll Settings**: anonymous, public-add-option\n**Total votes**: 0",
            	            	+  Text: (string) (len=70) "---\n**Poll Settings**: anonymous, public-add-option\n**Total Votes**: 0",
            	            	   Fields: ([]*model.SlackAttachmentField) <nil>,
            	Test:       	TestPollToPostActions/Multipile_questions,_settings:_anonymous,_public-add-option
```

@kaakaa Do you have any idea why this is happening?

kaakaa:
Thanks @hanzei! I'll take a look into new code later.

> @kaakaa Do you have any idea why this is happening?

You should fix the message from `**Total Votes**` to `**Total votes**`.
After fixing it, some tests failed by mismatch of `customize`  and `customise`. I think that We should use `customize` and `custmization`, because[ mattermost-server uses `Customization` ](https://github.com/mattermost/mattermost-server/blob/44887a0272ef17b5db8d64e80fd7fd0bafaa5080/model/config.go#L1599).

hanzei:
:man_facepalming: Thanks, fixed.

server/plugin/api_test.go
```diff
@@ -144,12 +145,14 @@ func TestServeFile(t *testing.T) {
 }
 
 func TestHandleVote(t *testing.T) {
+	localizer := i18n.NewLocalizer(&i18n.Bundle{})
```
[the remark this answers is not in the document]
hanzei: Good catch :+1:

hanzei:
I strongly suppose this issue is caused by a bug in the server. Reported at https://community-daily.mattermost.com/core/pl/9oyf339pajfx3msy1t14fr9poh.

kaakaa:
> I strongly suppose this issue is caused by a bug in the server. Reported at https://community-daily.mattermost.com/core/pl/9oyf339pajfx3msy1t14fr9poh.

I see, thanks.
For now, I can approve this PR, but are you intended to add commits for using `yaml`?

hanzei:
>  For now, I can approve this PR, but are you intended to add commits for using yaml?

I would opt to move this discussion into a separate issue. Are you fine with this?

kaakaa:
Sure! Thank you for your great work!

hanzei:
@kaakaa I've moved `i18n` into `assets` to streamline the process. Hope you are fine with this.

i18n/active.de.json
hanzei: See #139 for the follow up ticket

PR #137, opened 2019-03-19
 Update dependencies & downstream sample-plugin

hanzei:
This PR does a couple of things:
- Unity asset handling as in https://github.com/mattermost/mattermost-plugin-sample/pull/28
- Downstream change in `Makefile`
- Move the logo in `images` as it's not required be in the bundle
- Make use of `GetBundlePath()` introduced in https://github.com/mattermost/mattermost-server/pull/10466
- Update `mattermost-server` to latest `master`

Because of this changes the next version will require Mattermost v5.10. Do you see a problem with this?


hanzei:
> Could you fix this warning?

Surely, good catch. Fixed.

> Before merging, please set `minimumServerVersion` to `5.10.0`.

I would like to do, but first `5.10.0` must be added in https://github.com/mattermost/mattermost-server/blob/master/model/version.go. Do you want to defer this PR until then?

kaakaa:
Thank you for fixing the warning from vet!

> I would like to do, but first 5.10.0 must be added in https://github.com/mattermost/mattermost-server/blob/master/model/version.go. Do you want to defer this PR until then?

Sure, I got it 👍

PR #140, opened 2019-03-20
Use Bot Account

### LINKED ISSUE #126: Use Bot Account
#### Summary

5.8 will introduce Bot Accounts to Mattermost. We should make use of this. The changes I propose are:
- Create a Bot Account "Matterpoll" on activation 
- Store the userId in the KV Store
- Create all posts with this user.

#### Additional context

Link to spec: https://docs.google.com/document/d/139dNmjHz0Tflcgby85xaJtxYwapPctZHiZpZT6aS0DA

hanzei:
This PR makes use of the Bot Account feature coming in Mattermost v5.10.

Basically ever persistent post is now posted by the bot account. ~~Ephemeral messages are not, but this is because posting a ephemeral message as a bot account isn't supported at the moment.~~ This will be fixed with MM v5.12.

~~Please not that there are some bug in the server part. See https://community-daily.mattermost.com/core/pl/9oyf339pajfx3msy1t14fr9poh. Blocked until the webapp part of the feature is fully implemented.~~ All bugs are fixed!

The only "not bot" messages are the ones created in `api.go`. I will change this is a different PR.

Fixes #126

### DIFF
--- a/server/plugin/plugin.go
+++ b/server/plugin/plugin.go
@@ -2,6 +2,8 @@ package plugin
 
 import (
 	"fmt"
+	"io/ioutil"
+	"path/filepath"
 	"sync"
 
 	"github.com/blang/semver"
@@ -18,8 +20,13 @@ import (
 // MatterpollPlugin is the object to run the plugin
 type MatterpollPlugin struct {
 	plugin.MattermostPlugin
-	router *mux.Router
-	Store  store.Store
+	botUserID string
+	bundle    *i18n.Bundle
+	router    *mux.Router
+	Store     store.Store
+
+	// activated is used to track whether or not OnActivate has initialized the plugin state.
+	activated bool
 
 	// configurationLock synchronizes access to the configuration.
 	configurationLock sync.RWMutex
@@ -28,15 +35,24 @@ type MatterpollPlugin struct {
 	// setConfiguration for usage.
 	configuration *configuration
 	ServerConfig  *model.Config
-
-	bundle *i18n.Bundle
 }
 
-const minimumServerVersion = "5.10.0"
+var botDescription = &i18n.Message{
+	ID:    "bot.description",
+	Other: "Poll Bot",
+}
+
+const (
+	minimumServerVersion = "5.10.0"
+
+	botUserName    = "matterpoll"
+	botDisplayName = "Matterpoll"
+)
 
 // OnActivate ensures a configuration is set and initializes the API
 func (p *MatterpollPlugin) OnActivate() error {
-	if err := p.checkServerVersion(); err != nil {
+	var err error
+	if err = p.checkServerVersion(); err != nil {
 		return err
 	}
 
@@ -44,19 +60,37 @@ func (p *MatterpollPlugin) OnActivate() error {
 		return errors.New("siteURL is not set. Please set a siteURL and restart the plugin")
 	}
 
-	store, err := kvstore.NewStore(p.API, PluginVersion)
+	p.Store, err = kvstore.NewStore(p.API, PluginVersion)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "failed to create store")
 	}
-	p.Store = store
 
-	p.router = p.InitAPI()
-
-	bundle, err := p.initBundle()
+	p.bundle, err = p.initBundle()
 	if err != nil {
 		return errors.Wrap(err, "failed to init localisation bundle")
 	}
-	p.bundle = bundle
+
+	bot := &model.Bot{
+		Username:    botUserName,
+		DisplayName: botDisplayName,
+	}
+	botUserID, appErr := p.Helpers.EnsureBot(bot)
+	if appErr != nil {
+		return errors.Wrap(appErr, "failed to ensure bot user")
+	}
+	p.botUserID = botUserID
+
+	if err = p.patchBotDescription(); err != nil {
+		return errors.Wrap(err, "failed to patch bot description")
+	}
+
+	if err = p.setProfileImage(); err != nil {
+		return errors.Wrap(err, "failed to set profile image")
+	}
+
+	p.router = p.InitAPI()
+
+	p.setActivated(true)
 
 	return nil
 }
@@ -68,6 +102,14 @@ func (p *MatterpollPlugin) OnDeactivate() error {
 		return errors.Wrap(err, "failed to dectivate command")
 	}
 	return nil
+}
+
+func (p *MatterpollPlugin) setActivated(activated bool) {
+	p.activated = activated
+}
+
+func (p *MatterpollPlugin) isActivated() bool {
+	return p.activated
 }
 
 // checkServerVersion checks Mattermost Server has at least the required version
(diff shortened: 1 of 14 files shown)


server/plugin/plugin.go
```diff
@@ -81,6 +87,48 @@ func (p *MatterpollPlugin) checkServerVersion() error {
...
+			Description: "Some description",
+		}
+
+		rbot, appErr := p.API.CreateBot(bot)
+		if appErr != nil {
+			return errors.Wrap(appErr, "failed to create bot account")
+		}
+		botUserID = rbot.UserId
+
+		if appErr := p.API.SetProfileImage(botUserID, profileImage); appErr != nil {
```
hanzei: I'm a bit unhappy with this. If `SetProfileImage` fails, the bot account will never have a profile image. But I also don't want to set the image on every startup.

server/plugin/plugin.go
```diff
@@ -81,6 +87,48 @@ func (p *MatterpollPlugin) checkServerVersion() error {
...
+
+		profileImage, err := ioutil.ReadFile(filepath.Join(bundlePath, "assets", "logo_dark.png"))
+		if err != nil {
+			return errors.Wrap(err, "failed to read profile image")
+		}
+
+		bot := &model.Bot{
+			Username:    "matterpoll",
+			DisplayName: "Matterpoll",
+			Description: "Some description",
```
hanzei: Should this be localizes? How do we deal with changes of the default server language?

server/plugin/plugin_test.go
```diff
@@ -115,23 +116,15 @@ func TestPluginOnActivate(t *testing.T) {
...
-		/*
-			"GetBundlePath fails": {
-				SetupAPI: func(api *plugintest.API) *plugintest.API {
-					api.On("GetServerVersion").Return(minimumServerVersion)
-					api.On("GetBundlePath").Return("", errors.New(""))
-					return api
-				},
-				ShouldError: true,
-			},
-		*/
```
hanzei: Forgot to remove this :man_shrugging:

server/plugin/plugin_test.go
```diff
@@ -115,23 +116,15 @@ func TestPluginOnActivate(t *testing.T) {
...
 	} {
 		t.Run(name, func(t *testing.T) {
 			api := test.SetupAPI(&plugintest.API{})
 			defer api.AssertExpectations(t)
 
 			patch := monkey.Patch(kvstore.NewStore, func(plugin.API, string) (store.Store, error) {
-				return &mockstore.Store{}, nil
+				store := &mockstore.Store{}
+				store.BotStore.On("GetID").Return(testutils.GetBotUserID(), nil)
```
hanzei: A bit ugly, but I don't want to refactor everything.

server/plugin/plugin.go
```diff
@@ -81,6 +87,48 @@ func (p *MatterpollPlugin) checkServerVersion() error {
 	return nil
 }
 
+// ensureBotAccount checks if the bot account exists and creates him if doesn't
+func (p *MatterpollPlugin) ensureBotAccount() error {
+	botUserID, err := p.Store.Bot().GetID()
+	if err != nil {
+		return errors.Wrap(err, "failed to get bot user from store")
+	}
```
hanzei: What should happen, if an admin disables the bot? Just enable it back?

server/plugin/plugin.go
hanzei: Ideas for the wording?

server/plugin/plugin.go
kaakaa: Hmm...I'm not too clear about that point.
If `SetProfileImage` fails, this func returns error, so plugin never be started. Bots without images will never be registered. Is this right?

server/plugin/plugin.go
hanzei: Maybe this isn't a problem. I've tested and you can still create messages, if the bot account is disabled.

server/plugin/plugin.go
hanzei: Yes, users can see it. The description is shown in the profile pop up. 
Screenshot from 2019-03-24 19-48-35


:+1: on localizing this. The remaining questions is:

> How do we deal with changes of the default server language?

hanzei: FYI: The NPS plugins uses `Created by the Net Promoter Score plugin.` (https://github.com/mattermost/mattermost-plugin-nps/blob/master/server/activate.go#L69)

But I think I like `Poll plugin for Mattermost` more. Or `Poll Bot`?

server/plugin/plugin.go
hanzei: > Bots without images will never be registered. Is this right?

No. The account will be crated in L113.

kaakaa: Yes, Bot account is created, but the account will not be registered to KVStore, because this func return error before calling L123. A zombie bot account still alive.

So I suppose that Bots without images will never be registered. I think that this is not problem for now.

server/plugin/plugin.go
kaakaa: Thanks, I got it.
We can catch changes of server config by `OnConfigurationChage` hooks, and update the description of bot by `Bot.Patch` in that hook func.
Existing poll will not be updated, but it's fine for me.

kaakaa: `Poll Bot` sounds best for me 👍

server/plugin/plugin.go
kaakaa: It sounds good, but who does create a poll if bot is disabled?

kaakaa:
@hanzei Sorry for my super late reply, because March is the end of fiscal year in Japan. But now I'm back, and leave some comments. Please take a look.

hanzei:
@kaakaa No problem, I'm not in a hurry.

I will think a bit about this PR and then update the PR.

server/plugin/plugin.go
hanzei: I will change the code to always set the profile image.

hanzei:
I've updated the PR. The logic now is:

- Try to get the stored botID
- If this fails, try to create a new bot, after checking if one with the same username exists
- Then patch the description
- Then patch the profile image


@kaakaa Please review. If you think the logic is fine, I will add more tests.

server/plugin/plugin.go
```diff
@@ -131,5 +221,8 @@ func (p *MatterpollPlugin) SendEphemeralPost(channelID, userID, message string)
 	ephemeralPost.AddProp("override_username", responseUsername)
 	ephemeralPost.AddProp("override_icon_url", fmt.Sprintf(responseIconURL, *p.ServerConfig.ServiceSettings.SiteURL, PluginId))
 	ephemeralPost.AddProp("from_webhook", "true")
-	_ = p.API.SendEphemeralPost(userID, ephemeralPost)
+	err := p.API.SendEphemeralPost(userID, ephemeralPost)
```
kaakaa: `SendEphemeralPost` returns `*model.Post`, so this error check isn't needed.
https://developers.mattermost.com/extend/plugins/server/reference/#API.SendEphemeralPost

Sorry, I missed this.

hanzei: Fixed

hanzei:
@kaakaa I did a lot of changes. The big one being the use of `helpers.EnsureBot()` instead of my self-made solution for storing the bot information. I also had to include the concept of `p.activated` from #155 in this PR.

Please take an in depth look.

server/plugin/plugin.go
```diff
@@ -28,35 +35,62 @@ type MatterpollPlugin struct {
...
+		return errors.Wrap(err, "failed to patch bot description")
+	}
+
+	if err = p.setProfileImage(); err != nil {
+		return errors.Wrap(err, "failed to set profile image")
+	}
+
+	p.router = p.InitAPI()
+
+	p.setActivated(true)
```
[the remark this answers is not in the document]
hanzei: Fixed

Issue #138, opened 2019-03-20
Add docs on how to do translations


hanzei:
There should be `contribution.md` file that explains how to do translations and how to submit PR's for this.

Follow up on https://github.com/matterpoll/matterpoll/pull/133

Blocked by #139

hanzei:
For a proposal on the process see https://github.com/matterpoll/matterpoll/pull/145#issuecomment-480807477

PR #145, opened 2019-04-08
Add FR translation

ldidry:


hanzei:
Thanks for the PR @ldidry!

Would your also be open maintaining the France translation in the future?

ldidry:
Sure.

ldidry:
Could you mention me when there will be new strings to translate?

hanzei:
Sure. I'm not sure how the process will be (see https://github.com/matterpoll/matterpoll/issues/138). What you think about:

- When a new (major) version will releases soon, I will open an issue informing about this.
- I will ping all translators to inform them about this.
- They open PR's with new translations, which (may) get reviewed by other translators.
- After all translation PR's are merged, the new version is released.

ldidry:
Perfect :ok_hand:

PR #146, opened 2019-04-15
Fix NPE in addOption

hanzei:
If `request == nil` we can't use the information to send a ephemeral post. Since we don't have any information, there is nothing we can do here other then logging the error, right?


hanzei:
Found by https://golangci.com/ :wink:

kaakaa:
wow, this is perfect sales promotion 😆

PR #149, opened 2019-04-22
Add japanese translation

kaakaa:


assets/i18n/active.ja.json
```diff
@@ -0,0 +1,122 @@
...
+    "hash": "sha1-390ff4fdd0cfef6ef35fbf067a605aa3ea87ecca",
+    "other": "回答を削除する"
+  },
+  "poll.button.endPoll": {
+    "hash": "sha1-60ddde386bd7190fa3da2ba13f97e3caa625d3e6",
+    "other": "投票を終了する"
+  },
+  "poll.endPost.answer.heading": {
+    "hash": "sha1-46743e9dbbeec0c43d16fe387735cf95c2e27abb",
+    "other": "{{.Answer}} (投票数 {{.Count}})"
```
hanzei: Just to double check: There is no difference between the number of votes in the wording?

kaakaa: Yes, Japanese has lesser cases for using plural than English, so I think this message would not be wrong.
(FYI: http://selftaughtjapanese.com/2014/02/05/plurality-in-japanese/)

Thanks for taking look deeply 👍

PR #152, opened 2019-04-24
Add 1.1.0 docs

hanzei:


README.md
```diff
@@ -30,6 +33,17 @@ If you want to define all answer options by yourself, type `/poll "Is Matterpoll
 Poll Settings provider further customisation, e.g. `/poll "Is Matterpoll great?" "Of course" "In any case" "Definitely" --progress --anonymous`. The available Poll Settings are:
 - `--anonymous`: Don't show who voted for what at the end
 - `--progress`: During the poll, show how many votes each answer option got
+- `--public-add-option`: Allow all users to add additional options
+
+
+## Localization
```
kaakaa: It would be better to write how to change language of matterpoll.

For example, 

```
You can change language of poll message by setting it in **System Console > General > Localization > Default Server Language**.
Language of messages that only you can see (e.g.: help messages, error messages) can be set from **Account Settings > Display > Language**.
```

How do you feel about adding this lines to the last of `Localization` section?

hanzei: Good idea. I've updated the text.

PR #155, opened 2019-04-29
Add translation for slash command

### LINKED ISSUE #154: Missing translation for slash command
#### Summary
there is no translationentry for some content

#### Steps to reproduce
just tap `/poll` in the client side and the autoComplete can't be translate

#### Code impact

begin line 156 :
```go
func getCommand(trigger string) *model.Command {
	return &model.Command{
		Trigger:          trigger,
		DisplayName:      "Matterpoll",
		Description:      "Polling feature by https://github.com/matterpoll/matterpoll",
		AutoComplete:     true,
		AutoCompleteDesc: "Create a poll",   <<<< **this one**
		AutoCompleteHint: `"[Question]" "[Answer 1]" "[Answer 2]"...`, <<<< **and this one**
	}
}
```

hanzei:
This PR adds support for translating the slash command help text.

Fixes #154


server/plugin/configuration.go
```diff
@@ -24,18 +25,19 @@ func (p *MatterpollPlugin) OnConfigurationChange() error {
 		return errors.New("Empty trigger not allowed")
 	}
 
-	if oldConfiguration.Trigger != "" {
-		if err := p.API.UnregisterCommand("", oldConfiguration.Trigger); err != nil {
-			return errors.Wrap(err, "failed to unregister old command")
+	// Don't try to register a command if the plugin isn't activated
+	if p.isActivated() {
```
hanzei: I had to add this because `OnConfigurationChange()` is called before `OnActivate()`, but then `p.bundle` is `nil`.

server/plugin/plugin.go
```diff
@@ -30,6 +30,9 @@ type MatterpollPlugin struct {
 	ServerConfig  *model.Config
 
 	bundle *i18n.Bundle
+
+	// activated is used to track whether or not OnActivate has initialized the plugin state.
+	activated bool
```
hanzei: While I don't like this, I didn't see a better solution.

server/plugin/plugin_test.go
```diff
@@ -20,17 +20,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func setupTestPlugin(t *testing.T, api *plugintest.API, store *mockstore.Store, siteURL string) *MatterpollPlugin {
-	defaultServerLocale := "en"
+func setupTestPlugin(t *testing.T, api *plugintest.API, store *mockstore.Store) *MatterpollPlugin {
```
hanzei: I don't know why we would pass a `siteURL` here. Removing it makes tests much easier.

server/plugin/configuration.go
kaakaa: Can we call `initBundle` in `OnConfigurationChange`? If so, we can remove `if p.isActivated()`.

hanzei: Yes, we could do this. But it feels more hacky to me. 2/5

kaakaa: As you said my suggestion looks hacky, but I don't have a better idea for now.
It's best not to call `OnConfigurationChange` before calling `OnActivate`, however it would be difficult.

So I'm fine to go with this code.

Issue #28, opened 2018-08-24
Change the look of voted answer


kaakaa:
We cannot see on browser which answers I voted.
So the look of voted answer should be changed. (e.g: Change background color, Add (*) as prefix, ...)

Sample image: 
![2018-08-24 13 22 52](https://user-images.githubusercontent.com/1453749/44565126-894e8d80-a7a1-11e8-99e2-e27c92571169.png)

This feature may need to implement webapp plugin, because UI changes are different for each user.

Hanzei:
I like the idea but I'm not sure if its possible to create this. Its not possible to change the text or color for the server side. It might work with webapp but I'm sceptical.

kaakaa:
Yes, I don't know how can we implement this now. But I think this feature is necessary.
I'll investigate more about webapp plugin.

hanzei:
Since https://github.com/mattermost/mattermost-server/pull/9859 got merged: Are there more changes in Mattermost required?

kaakaa:
Thank you for letting me know it!
I don't think that more changes in Mattermost is needed.

And now, [my super early prototype](https://github.com/matterpoll/matterpoll/commit/08bc956b27d77fce8beb68c4bc0d97311d3d1662) can change style of voted answer.
The followings are things to consider.

### Architecture
For achiving this feature, I'll add `answers` props to the post and parse it on webapp plugin.

```
"props": {
    "attachments": {...},
    "answers": {
        "answer_option1": ["user_id1"],
        "answer_option2": ["user_id2", "user_id3"],
        ...
    }
}
```

### Anonymous Settings
Since all clients receive `answers` props, anonymous setting has less meanings.
Of course, `answers` info isn't displayed, but I think that users can get `answers` info with modifying js.

### AnswerOptionId
Now, the voted answer is distinguished by answer string. So if answer strings are duplicated, Matterpoll can't change the style of buttons properly.
We should add OptionId to AnswerOption or reject poll that has duplicated answer string.



How do these sounds for you?

hanzei:
First of all the prototype is awesome!  :tada: 

#### Architecture

We could also add a new API endpoint and fetch this information via http. This might be a cleaner approach. 

#### Anonymous Settings

This might be a concern. The APi approach would help here.

> Now, the voted answer is distinguished by answer string. So if answer strings are duplicated, Matterpoll can't change the style of buttons properly.
We should add OptionId to AnswerOption or reject poll that has duplicated answer string.

[...]

PR #160, opened 2019-05-30
Improve poll UI

kaakaa:
## Summary
I work on #28 again. (Sorry to take so long to do this).

スクリーンショット 2019-05-30 23 38 34

This PR would work fine, but There is still a problem.

## Concerns

For now, many of webapp code is almost the same as the corresponding part of mattermost-webapp, in order to work in the same way as the interactive button on mattermost-webapp.
So we will have to follow the upstream, and I don't know how hard the work is.

## Proposed
So I propose that we release this feature as **experimental feature** at first.
Luckily, Enabled / disabled can be switched by just toggling `model.Post.Type` between `"custom_matterpoll"` and `model.POST_DEFAULT`.
I supposed that it can be realized by adding `Rich Poll UI` setting to plugin settings. Users can use rich UI by setting `Rich Poll UI` setting to `true`.

What do you this about this? @hanzei 
If you are fine, I'll work for implementing it and adding test code.

Thanks.


server/plugin/api.go
```diff
@@ -297,6 +312,40 @@ func (p *MatterpollPlugin) handleAddOptionDialogRequest(w http.ResponseWriter, r
...
+
+	b, err := json.Marshal(v)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		p.API.LogError(fmt.Sprintf("Failed to marshal voted answers: %v", v))
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Write(b)
```
golangcibot: Error return value of `w.Write` is not checked (from `errcheck`)

hanzei:
Hey @kaakaa,

First of all: This is amazing and greatly improves the usability of Matterpoll!

I have two concerns to add myself:
- I can't review webapp code. As you probably know: I don't know react nor js. Hence, I can't review this PR and others that will follow. I currently also don't have the time to learn a new language. This worries me as a maintainer.
- Different UX between webapp and mobile: Currently plugins can only alter the webapp and not the mobile app. Because of this feature like this won't be available on mobile. This can lead to confusion.

But I also think that we need some of the feature only the webapp provides.

About your proposal: The idea of adding a setting for this is good :+1:.
I would like to add `custom_matterpoll` to all future posts make the update path easier. If we decide to remove the setting and enable this by default, also past posts would get the new UI. I would much rather only render the new post type, if the setting is enabled. Do you think this is feasible?

kaakaa:
Thank you for commenting @hanzei 

* I'm also not professional about React/Redux, and still learning it now. It is the strong reason why I would like to release this feature as experimental. And I can understand your situation, you work hard. How about if we ask community to work as reviewer? 
* Yeah, mobile app has problems... This feature might be acceptable, but some forthcoming features such as multiple answers might not be endured. Do you have any info about mobile plugin?

> I would like to add custom_matterpoll to all future posts make the update path easier. If we decide to remove the setting and enable this by default, also past posts would get the new UI. I would much rather only render the new post type, if the setting is enabled. Do you think this is feasible?

The update path you suggested makes sense and is possible, but I think it's not feasible. Because post 
 with `custom_matterpoll` must be rendered by webapp plugin. And Mattermost doesn't publish their react component, so we have to implement another post component that will be rendered if experimental settings is off. 
If `custom_matterpoll` become default, UI for past post will not be changed, but poll works fine. So I think that it's not critial.

hanzei:
> How about if we ask community to work as reviewer?

Did you have someone special in mind?

hanzei:
> How about if we ask community to work as reviewer?

Did you have someone special in mind?

> Do you have any info about mobile plugin?

No. The only thing I know, is that people are starting to work on this. :disappointed: 

>  Because post with `custom_matterpoll` must be rendered by webapp plugin.

Otherwise the post will not get rendered at all?

> And Mattermost doesn't publish their react component, so we have to implement another post component that will be rendered if experimental settings is off.

So we can't so something like: If the experimental setting is on, use the default "Mattermost" component?

kaakaa:
> > How about if we ask community to work as reviewer?
> 
>  Did you have someone special in mind?

Sorry, there is no special one in my mind... 😔 

> No. The only thing I know, is that people are starting to work on this. 😞

Making thigs more worse, I know little about React-Native...  First off, I'll look into it.

> So we can't so something like: If the experimental setting is on, use the default "Mattermost" component?

I didn't know some specs of mattermost plugin. But I had read code of other plugins and docs of plugin, and I can implement what we want. With previous commit, changing plugin configuration from system console can be emitted to client in real time, and experimental UI can be switched.
So this problem would be resolved. Thanks for your comments 👍

hanzei:
I asked in the Extension Channel for help with the review: https://community-release.mattermost.com/core/pl/xyscr4zdejbc7p3o475stbpbca

>  With previous commit, changing plugin configuration from system console can be emitted to client in real time, and experimental UI can be switched.

This is amazing!

kaakaa:
Hi @hanzei.
This PR is set to review.  If you find something in server code or so on, please let me know.
And I ask community members to help our reviews again.

## demo
test

webapp/src/plugin.jsx
```diff
@@ -0,0 +1,33 @@
...
+import {websocketHasVoted} from 'actions/vote';
+
+import {id as pluginId} from './manifest';
+import reducer from './reducers';
+
+export default class MatterPollPlugin {
+    async initialize(registry, store) {
+        const data = await fetchPluginConfiguration()();
+        if (data && data.experimentalui) {
+            await store.dispatch(configurationChange(registry, store, data));
```
scottleedavis: I don't think await is needed here.  Is it intended to wait before the other registry functions can register?

webapp/src/utils/link_only_renderer.js
```diff
@@ -0,0 +1,27 @@
...
+        let outHref = href;
+
+        if (!getScheme(href)) {
+            outHref = `http://${outHref}`;
+        }
+
+        let output = `<a class="theme markdown__link" href="${outHref}" target="_blank"`;
+
+        if (title) {
+            output += ' title="' + title + '"';
```
scottleedavis: following the style of the surrounding strings in this file, could this be ` title="${title}"` instead?

webapp/src/components/post_type/action_view/action_view.jsx
```diff
@@ -0,0 +1,60 @@
...
+                return;
+            }
+            switch (action.type) {
+            case 'button':
+                content.push(
+                    <ActionButton
+                        key={action.id}
+                        action={action}
+                        postId={this.props.post.id}
+                        hasVoted={answers.voted_answers && answers.voted_answers.includes(action.name)}
```
scottleedavis: IE doesn't not support `includes`, if this is desired to be supported `indexOf` can be used instead.

scottleedavis: https://caniuse.com/#search=includes

webapp/src/components/post_type/action_view/action_button/action_button.jsx
```diff
@@ -0,0 +1,52 @@
...
+        actions: PropTypes.shape({
+            voteAnswer: PropTypes.func.isRequired,
+        }).isRequired,
+    }
+
+    handleAction = (e) => {
+        e.preventDefault();
+        const actionId = e.currentTarget.getAttribute('data-action-id');
+
+        this.props.actions.voteAnswer(
```
scottleedavis: I honestly didn't know a prop could be a func.  Very cool.

webapp/src/actions/vote.js
```diff
@@ -0,0 +1,38 @@
...
+    });
+};
+
+export const fetchVotedAnswers = (siteUrl, pollId) => async (dispatch) => {
+    if (typeof pollId === 'undefined' || pollId === '') {
+        return;
+    }
+
+    let url = siteUrl;
+    if (!url.endsWith('/')) {
```
scottleedavis: these lines could be replaced with `let url = siteUrl.replace(/\/?$/, '/');`

scottleedavis:
Overall I personally think this looks great.  :)

webapp/src/actions/vote.js
kaakaa: Looks smart 😍

webapp/src/components/post_type/action_view/action_view.jsx
kaakaa: I didn't know that, thanks!

webapp/src/plugin.jsx
kaakaa: Good catch! it might have been remained for past implementation, but now it isn't needed.

kaakaa:
@scottleedavis 
I'm appreciate your precious reviews!
I've fixed some points that you commented, so it's ready for second round.

webapp/src/actions/vote.js
```diff
@@ -0,0 +1,38 @@
...
+        data: {
+            user_id: data.user_id,
+            poll_id: data.poll_id,
+            voted_answers: data.voted_answers,
+        },
+    });
+};
+
+export const fetchVotedAnswers = (siteUrl, pollId) => async (dispatch) => {
+    if (typeof pollId === 'undefined' || pollId === '') {
```
scottleedavis: could instead be `if (Boolean(pollId))`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean

webapp/src/components/post_type/action_view/action_view.jsx
```diff
@@ -0,0 +1,60 @@
...
+        if (!actions || !actions.length) {
+            return '';
+        }
+
+        const content = [];
+        const votedAnswers = this.props.votedAnswers || {};
+        const answers = votedAnswers[this.props.post.props.poll_id] || {};
+
+        actions.forEach((action) => {
+            if (!action.id || !action.name) {
```
scottleedavis: Could possibly filter instead of a forEach with if statement.
e.g. `actions.filter( (action) => action.id && action.name ).forEach( (action) ...`

webapp/src/reducers/post_type.js
```diff
@@ -0,0 +1,18 @@
+import ActionTypes from 'action_types';
+
+export const postTypeComponent = (state = {}, action) => {
+    switch (action.type) {
+    case ActionTypes.REGIST_POST_TYPE_COMPONENT_ID:
+        if (action.data) {
+            const nextState = {...state};
+            if (action.data.postTypeComponentId == null) {
```
scottleedavis: consider `if (!action.data.postTypeComponentId)` or `if (Boolean(action.data.postTypeComponentId))`

webapp/src/utils/link_only_renderer.js
```diff
@@ -0,0 +1,27 @@
+import RemoveMarkdown from './remove_markdown';
+
+function getScheme(url) {
+    const match = (/([a-z0-9+.-]+):/i).exec(url);
```
scottleedavis: perhaps be explicit for http(s) scheme?
`/^(http:\/\/|https:\/\/)/`

server/plugin/api.go
```diff
@@ -311,6 +341,38 @@ func (p *MatterpollPlugin) handleAddOptionDialogRequest(vars map[string]string,
...
 
+func (p *MatterpollPlugin) handleUserVoted(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	pollID := vars["id"]
+	userID := r.Header.Get("Mattermost-User-Id")
+
+	poll, err := p.Store.Poll().Get(pollID)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		p.API.LogWarn("failed to write response", "error", err.Error())
```
hanzei: That error message is not correct.

server/plugin/api.go
```diff
@@ -311,6 +341,38 @@ func (p *MatterpollPlugin) handleAddOptionDialogRequest(vars map[string]string,
...
+		return
+	}
+
+	v, err := poll.GetVotedAnswer(userID)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		p.API.LogError(fmt.Sprintf("Failed to get voted answers with UserId: %s", userID))
```
hanzei:
```suggestion
		p.API.LogError("Failed to get voted answers", "userID", userID))
```

server/plugin/api.go
```diff
@@ -311,6 +341,38 @@ func (p *MatterpollPlugin) handleAddOptionDialogRequest(vars map[string]string,
...
+		return
+	}
+
+	b, err := json.Marshal(v)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		p.API.LogError(fmt.Sprintf("Failed to marshal voted answers: %v", v))
```
hanzei:
```suggestion
		p.API.LogError("Failed to marshal voted answers", "answers", v))
```

server/plugin/api_test.go
```diff
@@ -121,6 +121,49 @@ func TestServeFile(t *testing.T) {
 	}
 }
 
+func TestHandlePluginConfiguratoin(t *testing.T) {
```
hanzei:
```suggestion
func TestHandlePluginConfiguration(t *testing.T) {
```

server/poll/poll_test.go
```diff
@@ -239,6 +239,80 @@ func TestUpdateVote(t *testing.T) {
...
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+
+			answers, err := test.Poll.GetVotedAnswer(test.UserID)
+			if test.Error {
+				assert.NotNil(err)
+				return
+			}
```
hanzei: Nit: I would use an `else` here. 0/5

webapp/package.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "matterpoll",
+  "version": "1.1.0",
+  "description": "This plugin is webapp for matterpoll plugin.",
```
hanzei: Do we need a description?

webapp/src/actions/vote.js
kaakaa: I didn't know this func, thanks! However `Boolean` function was marked as not recommended by ESLint.
https://eslint.org/docs/2.0.0/rules/no-extra-boolean-cast

So I choose `if (!pollId)` simply. Are you fine with this?

webapp/src/components/post_type/action_view/action_view.jsx
kaakaa: It looks smart 👍

webapp/src/reducers/post_type.js
kaakaa: This intend to pass through empty string.
Because if user unregister custom post type component, the postTypeComponentId that is registered in state should be removed, and empty string will be set to `nextState.id`.

So `if (!action.data.postTypeComponentId)` and `if (Boolean(action.data.postTypeComponentId))` would be not valid here.

webapp/src/actions/vote.js
scottleedavis: Strange as I have used (and was recommend to) use Boolean in the mattermost-webapp which also uses ESlint.  I am OK with the simple (!pollId). :)

.gitignore
```diff
@@ -1,2 +1,3 @@
 dist
 assets/i18n/translate.*.*
+.vscode/
```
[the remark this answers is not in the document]
kaakaa: I sent this by mistake, good catching 👍

plugin.json
```diff
@@ -11,14 +11,26 @@
...
-    "settings": [{
-     "key": "Trigger",
-     "display_name": "Trigger Word",
-     "type": "text",
-     "help_text": "Trigger Word must be unique, and cannot begin with a slash or contain any spaces.",
-     "default": "poll"
-     }],
-     "footer": "* To report an issue, make a suggestion or a contribution, [check the repository](https://github.com/matterpoll/matterpoll)."
+    "settings": [
+      {
```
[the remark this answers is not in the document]
kaakaa: I think that the current format is correct. because RFC8259 shows example like this.

(Page 12) `This is a JSON array containing two objects`
https://www.rfc-editor.org/rfc/rfc8259

Is there other reasons to change?

webapp/package.json
kaakaa: This desc never be used, so should be removed. Thanks.

webapp/src/actions/vote.js
kaakaa: Ah...it may be caused by the difference of `.eslintrc.json`. It would be better to make settings of ESLink the same as mattermost-webapp's one.

However, it seems that a warning about Boolean function will occur even if the settings are the same as mattermost-webapp ... I'll take a look more.

webapp/src/utils/link_only_renderer.js
kaakaa: Sorry, I need a little more time to understand here. I'll comment later.

plugin.json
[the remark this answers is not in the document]
hanzei: Personal style :wink: I like to have this kind of code as small as possible. And I like it to be consistent with the go code.

1/5

webapp/src/actions/vote.js
kaakaa: I understand. ESLint reports `no-extra-boolean-cast` when using `Boolean` function in `if` condition.  So here `if (!pollId) {` looks good.

webapp/src/utils/link_only_renderer.js
kaakaa: Practically, `http` and `https` are enough here. However, Mattermost always create link for `ftp`, `tel` and `mailto`, and we can set arbitrary url schemes. 
https://docs.mattermost.com/administration/config-settings.html#custom-url-schemes

I think we should follow this specs here, so I want to leave this for now.
As you may know, existing code doesn't follow that specs and some bugs remain, so I'll fix this as another PR.

plugin.json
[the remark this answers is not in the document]
kaakaa: Got it. I wondered if there is a rule I do not know about the format of JSON.
I follow your style :+1:

Makefile
```diff
@@ -141,7 +141,7 @@ ifneq ($(HAS_SERVER),)
 	$(GO) test -race -gcflags=-l ./server/...
 endif
 ifneq ($(HAS_WEBAPP),)
-	cd webapp && $(NPM) run fix;
+	cd webapp && $(NPM) run fix && $(NPM) run test;
```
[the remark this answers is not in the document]
kaakaa: Created.
https://github.com/mattermost/mattermost-plugin-starter-template/pull/51

kaakaa:
Thank you so much @scottleedavis and @hanzei !
I've fixed all points you reviewed. If you have additional comments, I would be glad to hear it.

hanzei:
Will review tomorrow

webapp/src/utils/link_only_renderer.js
scottleedavis: Sounds good to me.

server/plugin/api.go
```diff
@@ -311,6 +341,38 @@ func (p *MatterpollPlugin) handleAddOptionDialogRequest(vars map[string]string,
...
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		p.API.LogWarn("failed to get poll", "error", err.Error())
+		return
+	}
+
+	v, err := poll.GetVotedAnswer(userID)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		p.API.LogError("Failed to get voted answers", "userID", userID)
```
hanzei: Would you please also log `err`?

server/plugin/api.go
```diff
@@ -311,6 +341,38 @@ func (p *MatterpollPlugin) handleAddOptionDialogRequest(vars map[string]string,
...
+		return
+	}
+
+	b, err := json.Marshal(v)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		p.API.LogError("Failed to marshal voted answers")
```
hanzei: Would you please also log `err`?

server/plugin/api_test.go
```diff
@@ -121,6 +121,49 @@ func TestServeFile(t *testing.T) {
...
+			defer api.AssertExpectations(t)
+			p := setupTestPlugin(t, api, &mockstore.Store{})
+
+			w := httptest.NewRecorder()
+			r := httptest.NewRequest(http.MethodGet, "/api/v1/configuration", nil)
+			r.Header.Add("Mattermost-User-ID", model.NewId())
+			p.ServeHTTP(nil, w, r)
+
+			result := w.Result()
+			require.NotNil(t, result)
```
hanzei: Nit: Please close the body
```suggestion
			require.NotNil(t, result)
			defer result.Body.Close()
```

server/plugin/api_test.go
```diff
@@ -508,6 +561,66 @@ func TestHandleAddOptionDialogRequest(t *testing.T) {
...
+			defer store.AssertExpectations(t)
+			p := setupTestPlugin(t, api, store)
+
+			w := httptest.NewRecorder()
+			r := httptest.NewRequest(http.MethodGet, fmt.Sprintf("/api/v1/polls/%s/voted", testutils.GetPollID()), nil)
+			r.Header.Add("Mattermost-User-ID", "userID1")
+			p.ServeHTTP(nil, w, r)
+
+			result := w.Result()
+			require.NotNil(t, result)
```
hanzei:
```suggestion
			require.NotNil(t, result)
			defer result.Body.Close()
```

server/plugin/api_test.go
```diff
@@ -508,6 +561,66 @@ func TestHandleAddOptionDialogRequest(t *testing.T) {
...
+
+			bodyBytes, err := ioutil.ReadAll(result.Body)
+			require.Nil(t, err)
+
+			assert.Equal(test.ExpectedStatusCode, result.StatusCode)
+			if test.ShouldError {
+				assert.Equal([]byte{}, bodyBytes)
+				assert.Equal(http.Header{}, result.Header)
+			} else {
+				assert.NotNil(bodyBytes)
```
hanzei: Would you please also check the content? 1/5 that this would be usefull

server/plugin/command.go
```diff
@@ -156,7 +156,10 @@ func (p *MatterpollPlugin) executeCommand(args *model.CommandArgs) (string, *mod
 		UserId:    p.botUserID,
 		ChannelId: args.ChannelId,
 		RootId:    args.RootId,
-		Type:      model.POST_DEFAULT,
+		Type:      MatterpollPostType,
+		Props: model.StringInterface{
```
hanzei: I'm 1/5 on using `map[string]interface{}` here and be explicit.

server/poll/poll_test.go
```diff
@@ -239,6 +239,79 @@ func TestUpdateVote(t *testing.T) {
 	}
 }
 
+func TestGetVotedAnswer(t *testing.T) {
+	for name, test := range map[string]struct {
+		Poll     poll.Poll
+		UserID   string
+		Error    bool
```
hanzei: Maybe stay consistent and use `ExpectedError`? 1/5

server/poll/poll_test.go
```diff
@@ -239,6 +239,79 @@ func TestUpdateVote(t *testing.T) {
...
+		Expected *poll.VotedAnswerResponse
```
hanzei: See above
```suggestion
		ExpectedResponse *poll.VotedAnswerResponse
```

server/poll/poll_test.go
```diff
@@ -239,6 +239,79 @@ func TestUpdateVote(t *testing.T) {
...
+			UserID: "",
+			Error:  true,
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+
+			answers, err := test.Poll.GetVotedAnswer(test.UserID)
+			if test.Error {
+				assert.NotNil(err)
```
hanzei: `answers` should be `nil` in this case.
```suggestion
				assert.NotNil(err)
				assert.Nil(answers)
```

webapp/.babelrc
```diff
@@ -0,0 +1,49 @@
+{
```
hanzei: I'm just wondering if the should/can stay consistent with https://github.com/mattermost/mattermost-plugin-starter-template/blob/master/webapp/.babelrc here?

webapp/src/actions/config.js
```diff
@@ -0,0 +1,31 @@
...
+    return async () => {
+        let data;
+        try {
+            data = await Client.getPluginConfiguration();
+        } catch (error) {
+            return error;
+        }
+        return data;
+    };
+};
```
hanzei: Nit: Please add an empty line to this and the other files.

webapp/src/plugin.jsx
```diff
@@ -0,0 +1,33 @@
+import {configurationChange, fetchPluginConfiguration} from 'actions/config';
+import {websocketHasVoted} from 'actions/vote';
+
+import {id as pluginId} from './manifest';
+import reducer from './reducers';
+
+export default class MatterPollPlugin {
+    async initialize(registry, store) {
+        const data = await fetchPluginConfiguration()();
```
hanzei: The problem with this is, that if a user is not logged in, the code still tries to fetch the data, which results in a `4ß3`. Would be great to fix this, but it can be done in a different PR. I don't no how much work this is.

webapp/webpack.config.js
```diff
@@ -0,0 +1,42 @@
+var path = require('path');
```
hanzei: You might want to sync this with https://github.com/mattermost/mattermost-plugin-starter-template/blob/master/webapp/webpack.config.js

hanzei:
I've noticed that the size of the buttons isn't the same on the two views. Can you fix this?
Rich UI: Screenshot from 2019-08-07 14-51-43
Normal: Screenshot from 2019-08-07 14-51-56

webapp/src/plugin.jsx
hanzei: Maybe https://github.com/mattermost/mattermost-plugin-nps/pull/50 helps?

webapp/.babelrc
kaakaa: You're right, and I missed. I've fixed 👍

webapp/src/plugin.jsx
kaakaa: Thanks for pointing. I'll take a look into it.

kaakaa:
> I've noticed that the size of the buttons isn't the same on the two views. Can you fix this?

Great catch! It's caused that react-bootstrap force to insert `margin-left: -5px`. I've fixed this by [this commit](https://github.com/matterpoll/matterpoll/pull/160/commits/5dfd1e74b285df4d6ca0267ab75ca18afcadfcdc#diff-fad6ef75f135ebf9fdd66cb1031ea3ea).

webapp/.babelrc
```diff
@@ -1,49 +1,4 @@
...
-            ],
-            plugins: [
-                '@babel/plugin-proposal-class-properties',
-                '@babel/plugin-syntax-dynamic-import',
-            ],
-        }
-    }
+    presets: ['@babel/preset-env', '@babel/preset-react'],
+    plugins: ['@babel/plugin-proposal-class-properties']
 }
```
hanzei: Please add a newline here

webapp/src/components/post_type/action_view/action_button/action_button.jsx
```diff
@@ -50,3 +52,7 @@ export default class ActionButton extends React.PureComponent {
         );
     }
 }
+
+const style = {
+    buttonToolbar: {'margin-left': 0},
+};
```
hanzei: Please add a newline here

hanzei:
> > I've noticed that the size of the buttons isn't the same on the two views. Can you fix this?
> 
> @hanzei Great catch! It's caused that react-bootstrap force to insert `margin-left: -5px`. I've fixed this by [this commit](https://github.com/matterpoll/matterpoll/pull/160/commits/5dfd1e74b285df4d6ca0267ab75ca18afcadfcdc#diff-fad6ef75f135ebf9fdd66cb1031ea3ea) (and [this additional commit](https://github.com/matterpoll/matterpoll/pull/160/commits/507f2ca2843c818a229d6a9c1ab1e6cd06099fac)).

Nice. It just caught my eye.

webapp/src/plugin.jsx
kaakaa: This problem could be resolved by [this commit](https://github.com/matterpoll/matterpoll/pull/160/commits/1b4cb034b5e9059a68c6ce9cb418f940586da019).

Although we can avoid the error by checking currentUserId, it will not fetch the configuration when logging in. So I think that we need to subscribe the websocket event that is published when logging in (it might be called `hello` event)

kaakaa:
I supposed that all points would be resolved. Please take a look again, thanks.

webapp/src/actions/vote.js
```diff
@@ -0,0 +1,35 @@
...
+    });
+};
+
+export const fetchVotedAnswers = (siteUrl, pollId) => async (dispatch) => {
+    if (!pollId) {
+        return;
+    }
+
+    let url = siteUrl.replace(/\/?$/, '/');
+    url = `${url}/plugins/${pluginId}/api/v1/polls/${pollId}/voted`;
```
hanzei: That is not correct. For some reason `url` ends with `/', hence `url` will be `http://localhost:8065//plugins...`. Can you please fix this?

kaakaa: Fixed. Good catch 👍

PR #161, opened 2019-05-30
Trim space when checking duplication

kaakaa:
Existing duplication check considers `No` and `No `(with trainling space) as different option.
It should be considered as same.

スクリーンショット 2019-05-30 23 53 47

### DIFF
--- a/server/poll/poll.go
+++ b/server/poll/poll.go
@@ -60,13 +60,14 @@ func NewPoll(creator, question string, answerOptions, settings []string) (*Poll,
 	return &p, nil
 }
 
-// AddAnswerOption adds a new aAnswerOption to a poll
+// AddAnswerOption adds a new AnswerOption to a poll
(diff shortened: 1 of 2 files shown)


kaakaa:
If so, we don't reject duplicated polls that differ only by leading/trailing spaces, don't we? 
This PR's objective is rejecting it.

hanzei:
It think we do. I would so something like:
```go
func (p *Poll) AddAnswerOption(newAnswerOption string) error {
	newAnswerOption = strings.TrimSpace(newAnswerOption)
	if newAnswerOption == "" {
		return errors.New("empty option not allowed")
	}
	for _, answerOption := range p.AnswerOptions {
		if answerOption.Answer == newAnswerOption {
			return fmt.Errorf("duplicate options: %s", newAnswerOption)
		}
	}
}
```

Wouldn't this solve the problem?

kaakaa:
Ah, I got it. Your code looks more clear than original. Thank you for elaborating!

PR #162, opened 2019-06-06
Refactor api.go

hanzei:
This PR refactors `api.go` to be much more DRY and unify handling of `PostActionIntegrationRequest` and `SubmitDialogRequest`. The only functional change is that error thrown from the http handlers will now be logged. The log level is `warning`.

### DIFF
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -11,6 +11,7 @@ import (
 	"github.com/mattermost/mattermost-server/model"
 	"github.com/mattermost/mattermost-server/plugin"
 	"github.com/nicksnyder/go-i18n/v2/i18n"
+	"github.com/pkg/errors"
 )
 
 const (
@@ -19,6 +20,11 @@ const (
 	iconFilename = "logo_dark.png"
 
 	addOptionKey = "answerOption"
+)
+
+type (
+	postActionHandler   func(map[string]string, *model.PostActionIntegrationRequest) (*i18n.Message, *model.Post, error)
+	submitDialogHandler func(map[string]string, *model.SubmitDialogRequest) (*i18n.Message, *model.SubmitDialogResponse, error)
 )
 
 var (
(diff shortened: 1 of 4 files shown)


server/plugin/api.go
```diff
@@ -110,101 +116,119 @@ func (p *MatterpollPlugin) handleLogo(w http.ResponseWriter, r *http.Request) {
...
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		if _, err = w.Write(response.ToJson()); err != nil {
+			p.API.LogWarn("failed to write PostActionIntegrationResponse", "error", err.Error())
+		}
 	}
+}
+
+func (p *MatterpollPlugin) handleVote(vars map[string]string, request *model.PostActionIntegrationRequest) (*i18n.Message, *model.Post, error) {
```
kaakaa: With this style, we can't use `Mattermost-User-Id` header.  So `postActionHandler` and `submitDialogHandler` should have `userId` or `http.Header` as arguments.

server/plugin/api.go
```diff
@@ -110,101 +116,119 @@ func (p *MatterpollPlugin) handleLogo(w http.ResponseWriter, r *http.Request) {
...
+
+		if msg != nil {
+			userLocalizer := p.getUserLocalizer(request.UserId)
+			p.SendEphemeralPost(request.ChannelId, request.UserId, p.LocalizeDefaultMessage(userLocalizer, msg))
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		if _, err = w.Write(response.ToJson()); err != nil {
+			p.API.LogWarn("failed to write PostActionIntegrationResponse", "error", err.Error())
```
kaakaa: `PostActionIntegrationResponse` -> `SubmitDialogRequest `

server/plugin/api.go
```diff
@@ -110,101 +116,119 @@ func (p *MatterpollPlugin) handleLogo(w http.ResponseWriter, r *http.Request) {
...
+		}
+
+		if msg != nil {
+			userLocalizer := p.getUserLocalizer(request.UserId)
+			p.SendEphemeralPost(request.ChannelId, request.UserId, p.LocalizeDefaultMessage(userLocalizer, msg))
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		if _, err = w.Write(response.ToJson()); err != nil {
```
kaakaa: If `response` is nil, app crash. We need to add nil check before this.

server/plugin/api.go
hanzei: I have thought about this and I have a solution that works with this code. Would you be fine if we discuss this in a separate PR?

server/plugin/api.go
hanzei: Good catch :+1:

server/plugin/api.go
kaakaa: Sure 👍

PR #163, opened 2019-06-13
Secure API request via Mattermost-User-ID header

hanzei:


server/plugin/api.go
```diff
@@ -116,7 +117,18 @@ func (p *MatterpollPlugin) handleLogo(w http.ResponseWriter, r *http.Request) {
...
 
-func (p *MatterpollPlugin) handlePostActionIntegrationRequest(handler postActionHandler) func(http.ResponseWriter, *http.Request) {
+func checkAuthenticity(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Header.Get("Mattermost-User-ID") == "" {
+			http.Error(w, "not authorized", http.StatusUnauthorized)
+			return
+		}
+
+		next.ServeHTTP(w, r)
```
hanzei: I love this solution.

server/plugin/api_test.go
```diff
@@ -632,7 +680,8 @@ func TestHandleEndPoll(t *testing.T) {
 			p := setupTestPlugin(t, api, store)
 
 			w := httptest.NewRecorder()
-			r := httptest.NewRequest("POST", fmt.Sprintf("/api/v1/polls/%s/end", testutils.GetPollID()), bytes.NewReader(test.Request.ToJson()))
+			r := httptest.NewRequest(http.MethodPost, fmt.Sprintf("/api/v1/polls/%s/end", testutils.GetPollID()), bytes.NewReader(test.Request.ToJson()))
```
kaakaa: `TestHandleAddOptionDialogRequest` doesn't have a test for `not-authorized`.   Are there any reasons?

server/plugin/api.go
```diff
@@ -85,13 +85,14 @@ func (p *MatterpollPlugin) InitAPI() *mux.Router {
 	r.HandleFunc("/"+iconFilename, p.handleLogo).Methods("GET")
```
kaakaa: Would you change `"GET"` to `http.MethodGet` for consistency?

server/plugin/api_test.go
hanzei: Nope, I just forgot it. Added a test for this.

server/plugin/api.go
hanzei: Good catch :+1:

PR #169, opened 2019-06-19
Add Circleci

hanzei:
Travis has been bought recently and is dying. To avoid being without a working CI one day, I propose to add circleci as a second CI for now and if it runs smoothly, get rid of travis.

I would like to use the orb from https://circleci.com/orbs/registry/orb/mattermost/plugin-ci, but this requires go modules and therefore #168.

The deployment to GitHub *should* work, but I've not run a successful test jet. I suppose it will work fine, once this is merged to `master`.


hanzei:
I've removed the release deployment from travis.

Let's hope the circle one works as expected :wink

PR #170, opened 2019-06-19
Check if http request can be made to the plugin

hanzei:
There has been a couple of issues in the past when TLS certificates where misconfigured or siteURL's invalid. This PR checks on startup if `$SITE_URL/plugins/com.github.matterpoll.matterpoll/` can be reached and logs an error if not.

@kaakaa Let me know if you think this is useful.

cc @scottleedavis: Maybe you have thoughts on this or find this useful for your own plugins.

Tests are currently missing. I can add them if this feature is appreciated.


kaakaa:
Yes, checking connections is useful. But, I think that checking by this PR would be flaky, because waiting 100ms may not be enough in some case.

How about adding a slash command to check connection? 

```
/matterpoll-test

(Create post if failing to connect to server)
Version: v1.2.0
Test Connection to server: FAILED
Details: "Your server is misconfigured. Please check if your siteURL is set and you TLS certificate is valid"
```

scottleedavis:
Having an extra command to test is nice however it will may result in an issue for problems still created instead of the sysadmin troubleshooting the issue first.

hanzei:
I don't like to the idea of exposing debug information to end users. If this test is flaky we could:
1. Increase the sleep time
2. Check X times if a connection can be made
3. Check if a connection to `$SITE_URL` can be made, which an be done without an own go routine.

kaakaa:
I see. I throw out my idea.

Of the three ideas you listed, I think `2.` is the best, because it would be able to fit to various environment. And I think that it's enough to retry about three times at one or few seconds intervals.

server/plugin/plugin.go
```diff
@@ -158,6 +168,27 @@ func (p *MatterpollPlugin) setProfileImage() error {
...
+// checkHTTPConnection checks if a http request to
+func (p *MatterpollPlugin) checkHTTPConnection() error {
+	siteURL := *p.ServerConfig.ServiceSettings.SiteURL
+	url := siteURL + "/plugins/" + manifest.Id + "/"
+
+	for i := 0; i < 3; i++ {
+		resp, err := http.Get(url)
+		if err != nil {
+			p.API.LogDebug("failed to make http request to root plugin URL", "URL", url, "error", err.Error())
+			time.Sleep(1 * time.Second)
```
kaakaa: Shouldn't it be retried when http.StatusNotFound is returned?
Because if sending the request before running plugin server, I supposed that server doesn't return  errors, and return `404 NotFound`.

hanzei: I agree that we should retry when `http.StatusNotFound` is returned. But shouldn't we also retry when an error is returned?

hanzei:
@kaakaa Do you have a good idea how to test this?

server/plugin/plugin.go
kaakaa: Since retry is for waiting for the server to start, I think the we don't need to retry when catching err. 1/5

server/plugin/plugin.go
```diff
@@ -90,6 +92,14 @@ func (p *MatterpollPlugin) OnActivate() error {
 
 	p.router = p.InitAPI()
 
+	go func(p *MatterpollPlugin) {
+		// Wait a bit to let the server mark the plugin as running
+		time.Sleep(100 * time.Millisecond)
+		if err = p.checkHTTPConnection(); err != nil {
+			p.API.LogError("Your server is misconfigured. Please check if your siteURL is set and you TLS certificate is valid", "error", err.Error())
```
kaakaa: nitpick: `you TLS` -> `your TLS`?

server/plugin/plugin.go
```diff
@@ -158,6 +168,36 @@ func (p *MatterpollPlugin) setProfileImage() error {
...
+// checkHTTPConnection checks if a http request to
+func (p *MatterpollPlugin) checkHTTPConnection() error {
+	siteURL := *p.ServerConfig.ServiceSettings.SiteURL
+	url := siteURL + "/plugins/" + manifest.ID + "/"
+
+	for i := 0; i < 3; i++ {
+		backOfTime := time.Duration(i) * 1 * time.Second
+		time.Sleep(backOfTime)
+
+		resp, err := http.Get(url)
```
kaakaa: I just noticed.
Here call `http.Get` simply, but if user is in company proxy environment, this might never reach to the host. Mattermost itself has code for proxy like [here](https://github.com/mattermost/mattermost-server/blob/dac7014b48218909c70f5b84e34b1f16f56bb17a/services/httpservice/client.go#L139)

I've encountered problems related to company proxy in may times, it needs huge effort to resolve this completely. Could we implement helper method like `ping` to mattermost-server?

hanzei:
Let's just postpone this PR. I'm not sure if this is a good idea. My concern is that there will be false positives.

hanzei:
@kaakaa What do you think about closing this in favour of https://mattermost.atlassian.net/browse/MM-14521?

kaakaa:
Yes, I think so. If reach to SiteURL is guaranteed, It seems we don't need to check `$SITE_URL/plugins/com.github.matterpoll.matterpoll/`.

We may need to take into account that starting matterpoll take too long time, but it won't be a problem of plugin.

hanzei:
Let's close this or now. We can reopen it, if mattermost.atlassian.net/browse/MM-14521 isn't working out so great.

PR #172, opened 2019-07-01

hanzei:
This PR downstreams the latest changes from https://github.com/mattermost/mattermost-plugin-starter-template. Notable are:
- Custom makefile changes are now in `build/custom.mk`
- `check-style` also checks for `golint` compliance
- Renamed `PollSettings` to `Settings`. I've tested with an older version and this isn't a breaking change. But please double check and test.


hanzei:
@kaakaa Did you also test the `PollSettings` change? I want to double check that this isn't a breaking change.

kaakaa:
@hanzei Yes, I did. I think its changes have no problem because it's just renaming type name. Are there any concerns?

hanzei:
No, just want to double check.

PR #188, opened 2019-08-07
Use bot account for ephemeral posts 

### LINKED ISSUE #187: Use bot accounts to post ephemeral responses
To make it look like this:

Screenshot from 2019-08-07 11-19-28

Follow up on #140

hanzei:
Fixes #187


server/plugin/api.go
```diff
@@ -143,10 +143,11 @@ func (p *MatterpollPlugin) handlePostActionIntegrationRequest(handler postAction
 			p.API.LogWarn("failed to handle PostActionIntegrationRequest", "error", err.Error())
 		}
 
-		response := &model.PostActionIntegrationResponse{}
 		if msg != nil {
-			response.EphemeralText = p.LocalizeDefaultMessage(userLocalizer, msg)
+			p.SendEphemeralPost(request.ChannelId, request.UserId, p.LocalizeDefaultMessage(userLocalizer, msg))
```
hanzei: This is the important change

kaakaa: Although I don't have any negative opinions for this changes, I don't understand importances for this fixes. Could you tell me any hints?

server/plugin/api_test.go
```diff
@@ -188,10 +190,11 @@ func TestHandleVote(t *testing.T) {
 				store.PollStore.On("Save", poll2Out).Return(nil)
 				return store
 			},
-			Request:            &model.PostActionIntegrationRequest{UserId: "userID1", PostId: "postID1"},
+			Request:            &model.PostActionIntegrationRequest{UserId: "userID1", ChannelId: "channelID1", PostId: "postID1"},
 			VoteIndex:          1,
 			ExpectedStatusCode: http.StatusOK,
-			ExpectedResponse:   &model.PostActionIntegrationResponse{EphemeralText: responseVoteUpdated.Other, Update: expectedPost2},
+			ExpectedMsg:        "Your vote has been updated.",
```
kaakaa: I think that using `responseVoteCounted.Other` is more stable than using text. 2/5

server/plugin/api.go
hanzei: Currently ephemeral messages are send `model.PostActionIntegrationResponse.EphemeralText`. The ephemeral message is posted as the user and the user icon and user name are overwritten to let it look like a Matterpoll message.

With this change we just use the bot account to create the message. This has the following advantages:
- We don't rely on overwriting the user icon
- The user icon is served (and cached) by the server, not by Matterpoll
- Clicking on the icon doesn't brings up the own users pop-up.

These aren't great advantages, but I think it's good be consistent and use the bot account everywhere.

server/plugin/api_test.go
hanzei: I'm been thinking about the same while rewriting the tests. I'm inclined to make tests more "stand alone". To me tests should not be written clever. They should be boring but easy to understand.

When I change something in the code I **want** to be forced to think about if this breaks a test. This is why I opted to use the explicit strings here.

Let me know what you think about this.

server/plugin/api.go
kaakaa: Thanks you for clarifying! I understand the importance of this change, and I agree with your thought :+1:. I also think that using bot account is better.

server/plugin/api_test.go
kaakaa: Tha followings are my thoughts for using variables instead of message text itself.

* Stable for changes of message text
  * No need to modify any tests when text changes
  * Tests don't test what text is returned, but test which message is returned, because all messages declared as variables
* Easy to search places where that message is used
  * Message text itself is like a magic number, so it should be avoided.

However, I also agree with your thoughts. Additionally, even if a message text is changed for now, fixing tests would be easy step because of little number of tests related to that text.

So let's go with this for now. thank you for explaining your thoughts.

PR #190, opened 2019-08-07
Rewrite DB upgrade

### LINKED ISSUE #189: Make DB upgrades easier
The current code for DB upgrades is very WET and hard to test. This should be improved.

hanzei:
This PR makes DB upgrades easier to test and write. It also makes the code more DRY.
Fixes #189


server/store/kvstore/upgrade.go
```diff
@@ -22,14 +34,26 @@ func (s *Store) UpdateDatabase(pluginVersion string) error {
 		return s.System().SaveVersion(newestSchema.String())
 	}
 
-	// TODO: Uncomment following condition when version 1.1.0 is released
-	/*
-		currentSchemaVersion := semver.MustParse(v)
-		if err := s.UpgradeDatabaseToVersion11(currentSchemaVersion); err != nil {
+	for _, upgrade := range s.upgrades {
+		v, err := s.System().GetVersion()
```
kaakaa: Do this need to call this on each iteration? Does it makes sense to update `v` after `SaveVersion` succeeds?

hanzei: Good catch :+1: I simplify the code to not fetch is on each iteration.

PR #191, opened 2019-08-09
MM-17366: handle curly quotes when parsing

lieut-data:
On iOS using an english keyboard, typing quotes yields curly quotes instead of straight quotes. Update the parser to handle these.

Fixes: https://mattermost.atlassian.net/browse/MM-17366


hanzei:
Thanks for the fix @lieut-data :+1:

PR #195, opened 2019-08-12
Add Golangci lint config

### LINKED ISSUE #194: Add golangci configuration file
Golangci can be configured via a config file. E.g. additional linter can be enabled or existing ones tweaked. Let's play a bit around and see if the find anything useful. See https://github.com/golangci/golangci-lint#configuration for the list of options.

hanzei:
I've played around a bit with the config options and found this mix quite well working. Feedback is highly appreciated :+1: 

Fixes  #194


server/poll/poll.go
```diff
@@ -54,7 +54,7 @@ func NewPoll(creator, question string, answerOptions, settings []string) (*Poll,
 		case "public-add-option":
 			p.Settings.PublicAddOption = true
 		default:
-			return nil, fmt.Errorf("Unrecognised poll setting %s", s)
+			return nil, fmt.Errorf("Unrecognized poll setting %s", s)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

hanzei: I'm aware of this issue, but it's not so easy to fix.

.golangci.yml
```diff
@@ -0,0 +1,24 @@
...
+    min-complexity: 20
+  gofmt:
+    simplify: true
+  golint:
+    # minimal confidence for issues, default is 0.8
+    min-confidence: 0
+  govet:
+    check-shadowing: true
+  misspell:
+    locale: US
```
hanzei: This is definitely debatable. I opted for `US` because the code pretty much didn't needed any changes. But `UK` is still an option.

server/poll/poll.go
hanzei: Let's fix this manually until #198 is merged.

.golangci.yml
kaakaa: Does using default setting make sense? Default setting is more loose than `US` or `UK`.

https://github.com/client9/misspell#usage
```
-locale string
    	Correct spellings using locale perferances for US or UK.  Default is to use a neutral variety of English.  Setting locale to US will correct the British spelling of 'colour' to 'color'
```

If we have to chose one of `US` or `UK`, I prefer `US`. 0/5

hanzei: Let's go with `US`.

.golangci.yml
```diff
@@ -0,0 +1,44 @@
...
+  maligned:
+    suggest-new: true
+  
+linters:
+  enable:
```
hanzei: I really want to add https://github.com/walle/lll to this list, but it's just to much work to fix all the long line for now.

hanzei:
@kaakaa Please take a second look

.golangci.yml
```diff
@@ -10,15 +15,30 @@ linters-settings:
...
     - golint
     - gosec
+    - maligned
     - misspell
+    - scopelint
     - unconvert
 
+issues:
+  exclude-rules:
+        # Exclude some linters from running on tests files.
```
kaakaa: super nitpick. this line seems to be wrong indent

.golangci.yml
kaakaa: It looks good. I think that long line looks ugly and may cause bugs because of the difficulty of understanding.

However, if we limit long line, are there any guidelines or rules to follow?
For example, how should we handle the following line?

```go
func (p *MatterpollPlugin) handleAddOptionDialogRequest(vars map[string]string, request *model.PostActionIntegrationRequest) (*i18n.Message, *model.Post, error) {
```

.golangci.yml
hanzei: Good catch, fixed

.golangci.yml
hanzei: I was wondering exactly the same :wink:

hanzei: Would you be fine if we most this discussion into a new issue/PR?

kaakaa: Sure!

PR #197, opened 2019-08-15
Add confimation dialog when deleting or ending a poll

### LINKED ISSUE #120: Add confimation dialog when deleting or ending a poll
#### Summary

We should open an interactive dialog when the user clicks on "End Poll" or "Delete Poll". Then a dialog pops up to confirm the action.

hanzei:
This PR adds a confirmation dialog when deleting or ending a poll. The code changes are pretty straight forward. I did some reordering on the code. Tests have been written.

I've also changed `PostEndPollAnnouncement` to log a warning instead of an error.

I'm 0/5 on the strings. Please double check and provide feedback.

Fixes #120


hanzei:
I just noticed that this is a breaking change for existing polls. Will need to fix this.

hanzei:
Hey @kaakaa, This is ready for review. The diff is quite big. Sorry for this. Will try to keep the footprint smaller next time.

server/plugin/api.go
```diff
@@ -299,7 +340,7 @@ func (p *MatterpollPlugin) handleAddOption(vars map[string]string, request *mode
 	return responseAddOptionSuccess, nil, nil
 }
 
-func (p *MatterpollPlugin) handleAddOptionDialogRequest(vars map[string]string, request *model.PostActionIntegrationRequest) (*i18n.Message, *model.Post, error) {
+func (p *MatterpollPlugin) handleEndPoll(vars map[string]string, request *model.PostActionIntegrationRequest) (*i18n.Message, *model.Post, error) {
```
golangcibot: 343-382 lines are duplicate of `server/plugin/api.go:452-492` (from `dupl`)

hanzei: Looks like this is our first false positive. I agree that the code is very similar, but I don't see a way of making this more DRY.

@kaakaa Thoughts on this?

server/plugin/api_test.go
```diff
@@ -381,6 +384,220 @@ func TestHandleVote(t *testing.T) {
...
+				api.On("OpenInteractiveDialog", dialogRequest).Return(nil)
+				return api
+			},
+			SetupStore: func(store *mockstore.Store) *mockstore.Store {
+				store.PollStore.On("Get", testutils.GetPollID()).Return(testutils.GetPollWithVotes(), nil)
+				return store
+			},
+			Request: &model.PostActionIntegrationRequest{
+				UserId:    "userID2",
+				PostId:    postID,
```
kaakaa: This successful test case need `ChannelId` field here, doesn't it?

server/plugin/api_test.go
```diff
@@ -1041,85 +1155,74 @@ func TestHandleEndPoll(t *testing.T) {
 				store.PollStore.On("Get", testutils.GetPollID()).Return(nil, &model.AppError{})
 				return store
 			},
-			Request:            &model.PostActionIntegrationRequest{UserId: "userID1", ChannelId: "channelID1", PostId: "postID1"},
+			Request:            &model.SubmitDialogRequest{UserId: "userID1", CallbackId: "postID1", TeamId: "teamID1"},
```
kaakaa: I want `ExpectedResponse: nil,` in this test case like others.

server/plugin/api.go
kaakaa: I don't think that we should extract common parts from these methods. Existing code is easy to understand because each steps is clear.

hanzei: Okay. I will disable `dupl` for now.

server/plugin/api.go
```diff
@@ -253,7 +242,59 @@ func (p *MatterpollPlugin) handleVote(vars map[string]string, request *model.Pos
...
+		}
+		if !hasPermission {
+			return responseAddOptionInvalidPermission, nil, nil
+		}
+	}
+
+	siteURL := *p.ServerConfig.ServiceSettings.SiteURL
+	dialog := model.OpenDialogRequest{
+		TriggerId: request.TriggerId,
+		URL:       fmt.Sprintf("%s/plugins/%s/api/v1/polls/%s/option/add", siteURL, manifest.ID, pollID),
```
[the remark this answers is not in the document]
hanzei: Are you sure we can? I thought the relative URLs can only be used for post actions.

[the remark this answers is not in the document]
kaakaa: Relative link for plugins is handled in [App.DoActionRequest](https://github.com/mattermost/mattermost-server/blob/master/app/integration_action.go#L210), and this URL is handled by it [here](https://github.com/mattermost/mattermost-server/blob/master/app/integration_action.go#L280). So relative links should woks fine. (I've tried it, and work fine)

Isn't this  allowed?

server/plugin/api_test.go
hanzei: Good catch :+1:

server/plugin/api_test.go
hanzei: You've got really good eyes :+1:

server/plugin/api.go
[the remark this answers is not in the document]
hanzei: You are 100% right. I don't know this was possible. Sorry for pushing back.

PR #198, opened 2019-08-15
Localize remaining strings

### LINKED ISSUE #196: Localize remaining strings
Despise our afford to localize all messages, two are still missing:
- Adding a already existing option to the poll shows `duplicate options`
- Creating a poll with an unknown setting e.g. with `/poll "test" --abc` shows `Unrecognised poll setting asd`

hanzei:
This PR localizes the last few remaining strings. It does that by adding a `poll.ErrorMessage` struct (0/5 on the naming). This message is meant for a user to consume and should not be wrapped.

I would love to hear feedback on the design.

Fixes #196


server/poll/poll.go
```diff
@@ -32,17 +32,24 @@ type Settings struct {
 	PublicAddOption bool
 }
 
-// NewPoll creates a new poll with the given paramatern
-func NewPoll(creator, question string, answerOptions, settings []string) (*Poll, error) {
+// ErrorMessage contains error messsage for a user that can be localized.
+// It should not be wrapped and instead allways returned.
```
kaakaa: `allways` -> `always`?

server/plugin/api.go
```diff
@@ -246,10 +246,15 @@ func (p *MatterpollPlugin) handleAddOption(vars map[string]string, request *mode
...
 	}
 
-	if err = poll.AddAnswerOption(answerOption); err != nil {
+	userLocalizer := p.getUserLocalizer(poll.Creator)
+
+	if errMsg := poll.AddAnswerOption(answerOption); errMsg != nil {
 		response := &model.SubmitDialogResponse{
 			Errors: map[string]string{
-				addOptionKey: err.Error(),
+				addOptionKey: p.LocalizeWithConfig(userLocalizer, &i18n.LocalizeConfig{
```
kaakaa: It seems more clearer by creating new func `LocalizeErrorMessageWithConfig(*i18n.Localizer, poll.ErrorMessage) string` that wraps `LocalizeWithConfig`.

hanzei: Good idea :+1:

server/plugin/command.go
```diff
@@ -121,18 +121,21 @@ func (p *MatterpollPlugin) executeCommand(args *model.CommandArgs) (string, *mod
...
+		newPoll, errMsg = poll.NewPoll(creatorID, q, o, s)
 	}
-	if err != nil {
+	if errMsg != nil {
 		appErr := &model.AppError{
 			Id: p.LocalizeWithConfig(userLocalizer, &i18n.LocalizeConfig{
 				DefaultMessage: commandErrorInvalidInput,
 				TemplateData: map[string]interface{}{
-					"Error": err.Error(),
+					"Error": p.LocalizeWithConfig(userLocalizer, &i18n.LocalizeConfig{
```
kaakaa: I missed to point out here. Here should also use `LocalizeErrorMessage `.

hanzei: Fixed this one

server/plugin/command.go
```diff
@@ -121,18 +121,18 @@ func (p *MatterpollPlugin) executeCommand(args *model.CommandArgs) (string, *mod
...
+		newPoll, errMsg = poll.NewPoll(creatorID, q, o, s)
 	}
-	if err != nil {
+	if errMsg != nil {
 		appErr := &model.AppError{
 			Id: p.LocalizeWithConfig(userLocalizer, &i18n.LocalizeConfig{
 				DefaultMessage: commandErrorInvalidInput,
 				TemplateData: map[string]interface{}{
-					"Error": err.Error(),
+					"Error": p.LocalizeErrorMessage(userLocalizer, errMsg),
```
hanzei: Just noticed that I'm wrapping an error here :man_shrugging:

kaakaa: I don't have a good idea for this...
Do you intend to do further changes for this? If not, I'll approve this PR.

hanzei: No, let's just go with this for now.

PR #207, opened 2019-08-19
Make webapp code private

kaakaa:
Make webapp code private, because webapp will never be published and should stay consiste with [mattermost-webapp](https://github.com/mattermost/mattermost-webapp/blob/master/package.json).

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -1,6 +1,7 @@
 {
   "name": "matterpoll",
-  "version": "1.1.0",
+  "version": "0.0.1",
+  "private": true,
   "main": "src/index.js",
   "scripts": {
     "build": "webpack --mode=production",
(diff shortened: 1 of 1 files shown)


webapp/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "matterpoll",
-  "version": "1.1.0",
+  "version": "0.0.1",
```
hanzei: What do you think about just removing the version number and rely on the version in `plugin.json`?

kaakaa: https://docs.npmjs.com/files/package.json#version
> If you don’t plan to publish your package, the name and version fields are optional.

Since `version` is optional, removing it is no problem.
Additionally, `0.0.1` has no means, so it should be removed. Thanks.

Issue #73, opened 2018-09-24
Cosmetic: Buttons "End poll" and "Delete Poll" only visible for creator


Twilek-de:
It would be nicer if only the poll creator would see the "End poll" and "Delete Poll" buttons. Maybe admins should have the power to end and delete polls as well...

Hanzei:
I don't think this is possible with the current Mattermost system. The problem is that every user sees the same message and therefore the same buttons.

@kaakaa Do you have any idea how to accomplish this?

Hanzei:
> Maybe admins should have the power to end and delete polls as well...

This is a nice idea. Would you open a new issue for this?

Twilek-de:
Done

kaakaa:
> @kaakaa Do you have any idea how to accomplish this?

Sorry, now I can't accomplish this.
But in terms of changing the display for each user, #28 is related to this issue. I supposed that it can be realized by making PostAttachment pluggable. But it is just an idea by now, and it will take more time to investigate and implement.

lxm065:
i think Poll Settings can provider further customisation for them,such as delete,end,visible and so on.

hanzei:
@lxm065 I'm not sure I'm getting your thoughts right. Do you want to add a Poll Setting for showing this buttons?

lxm065:
> @lxm065 I'm not sure I'm getting your thoughts right. Do you want to add a Poll Setting for showing this buttons?

yeah,such like /poll "it's all right?" "yes" "no"  --progress --visible 

by the way,is there any other api command?i find that hubot can't send slash commands
https://github.com/slackapi/hubot-slack/issues/202

hanzei:
> yeah,such like /poll "it's all right?" "yes" "no" --progress --visible


> by the way,is there any other api command?i find that hubot can't send slash commands
slackapi/hubot-slack#202

This is out of the scope of this issue. Could you please open a new one or post this question in the [Matterpoll Channel](https://community.mattermost.com/core/channels/matterpoll)?

hanzei:
> yeah,such like /poll "it's all right?" "yes" "no" --progress --visible

I see. But why would you allow everyone to end and delete a poll? I don't see the use case. 

> by the way,is there any other api command?i find that hubot can't send slash commands
slackapi/hubot-slack#202

This is out of the scope of this issue. Could you please open a new one or post this question in the [Matterpoll Channel](https://community.mattermost.com/core/channels/matterpoll)?x

[...]

PR #234, opened 2019-10-22
73 systembutton

kaakaa:
Refs #73

### Summary
This PR makes "End poll" and "Delete poll" button visible for only creator and system_administrator.

screenshot

### Note:
Now, "Add Option" button is not target of this PR, because the visibility of "Add Option" is differ from "End poll" and "Delete poll" because of `public-add-option` setting. I'll work on handling visibility of "Add Options" button after this PR is merged.


hanzei:
@kaakaa CI fails with:
```
server/poll/poll.go:36:6: type name will be used as poll.PollMetadata by other packages, and that stutters; consider calling this Metadata
Found 1 lint suggestions; failing.
```

hanzei:
@scottleedavis If you have the time to review this, it would be highly appreciated.

server/poll/poll.go
```diff
@@ -142,10 +143,20 @@ func (p *Poll) GetVotedAnswer(userID string) (*VotedAnswerResponse, error) {
...
 	}
-	return &VotedAnswerResponse{
-		PollID:       p.ID,
-		UserID:       userID,
-		VotedAnswers: votedAnswer,
+	return votedAnswer, nil
+}
+
+// GetMetadata returns personalized metadata of a poll
+func (p *Poll) GetMetadata(userID string, permission bool) (*Metadata, error) {
```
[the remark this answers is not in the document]
kaakaa: Thank you for interesting suggestion. But I think that pointer type is fine here. Because...

* Return value of `GetMetadata` is used just as the receiver of funcs ([`metadata.EncodeToByte()`](https://github.com/matterpoll/matterpoll/pull/234/files#diff-4444f0caa92e553b5c018c908e90e5e6R541) and [`metadata.ToMap()`](https://github.com/matterpoll/matterpoll/pull/234/files#diff-4444f0caa92e553b5c018c908e90e5e6R234)). So if changing return type of `GetMetadata` to value type, we should change receiver of these funcs. But, as far as I know, the receiver of these funcs should be pointer type.
* If return type of `GetMetadata` is value type, we must return [the empty instance of `Metadata`](https://github.com/matterpoll/matterpoll/pull/234/files#diff-00ec2aff510fa90953e4a6d689e108fdR153) when error occurs in `GetMetadata`. It looks little strange.

I know that returning the value type might be better than returning the pointer type ([ref1](https://medium.com/@philpearl/bad-go-pointer-returns-340f2da8289), [ref2](https://stackoverflow.com/questions/23542989/pointers-vs-values-in-parameters-and-return-values/23551970#23551970)), but I think that pointer type is fine for the above reasons. 1/5

Of course, I haven't looked up everything about pointer vs value type, so I'm welcoming any references about this problem.

server/plugin/api.go
```diff
@@ -219,15 +219,17 @@ func (p *MatterpollPlugin) handleVote(vars map[string]string, request *model.Pos
 		return commandErrorGeneric, nil, errors.Wrap(err, "failed to save poll")
 	}
 
-	v, err := poll.GetVotedAnswer(userID)
+	hasAdminPermission, appErr := p.HasAdminPermission(poll, userID)
+	if appErr != nil {
+		p.API.LogWarn("Failed to check permission", "userID", userID, "error", appErr.Error())
```
hanzei: Nit:
```suggestion
		p.API.LogWarn("Failed to check admin permission", "userID", userID, "error", appErr.Error())
```

server/plugin/api.go
```diff
@@ -219,15 +219,17 @@ func (p *MatterpollPlugin) handleVote(vars map[string]string, request *model.Pos
...
 	if err != nil {
 		return commandErrorGeneric, nil, errors.Wrap(err, "failed to get voted answers")
 	}
-	p.API.PublishWebSocketEvent("has_voted", map[string]interface{}{
-		"user_id":       v.UserID,
-		"poll_id":       v.PollID,
-		"voted_answers": v.VotedAnswers,
-	}, &model.WebsocketBroadcast{UserId: userID})
+
+	p.API.PublishWebSocketEvent("has_voted", metadata.ToMap(), &model.WebsocketBroadcast{UserId: userID})
```
hanzei: 0/5 Maybe make this a it's own method:
```golang
p.PublishPollMetadat(p *poll.Poll, userID string)
```
and move the code above also in this method

server/poll/poll.go
```diff
@@ -32,11 +32,12 @@ type Settings struct {
 	PublicAddOption bool
 }
 
-// VotedAnswerResponse stores answers that is created by a user
-type VotedAnswerResponse struct {
-	PollID       string   `json:"poll_id"`
-	UserID       string   `json:"user_id"`
-	VotedAnswers []string `json:"voted_answers"`
+// Metadata stores personaized metadata of a poll
+type Metadata struct {
```
hanzei: It would be great to have comment for all fields of the struct.

server/poll/poll.go
```diff
@@ -142,10 +143,20 @@ func (p *Poll) GetVotedAnswer(userID string) (*VotedAnswerResponse, error) {
...
 		}
 	}
-	return &VotedAnswerResponse{
-		PollID:       p.ID,
-		UserID:       userID,
-		VotedAnswers: votedAnswer,
+	return votedAnswer, nil
+}
+
+// GetMetadata returns personalized metadata of a poll
```
hanzei:
```suggestion
// GetMetadata returns personalized metadata of a poll.
```

server/poll/poll.go
```diff
@@ -32,11 +32,12 @@ type Settings struct {
 	PublicAddOption bool
 }
 
-// VotedAnswerResponse stores answers that is created by a user
-type VotedAnswerResponse struct {
-	PollID       string   `json:"poll_id"`
-	UserID       string   `json:"user_id"`
-	VotedAnswers []string `json:"voted_answers"`
+// Metadata stores personaized metadata of a poll
```
hanzei:
```suggestion
// Metadata stores personaized metadata of a poll.
```

server/poll/poll.go
```diff
@@ -129,8 +130,8 @@ func (p *Poll) UpdateVote(userID string, index int) error {
 	return nil
 }
 
-// GetVotedAnswer collect voted answers by a user and returns it as VotedAnswerResponse
-func (p *Poll) GetVotedAnswer(userID string) (*VotedAnswerResponse, error) {
+// getVotedAnswers collect voted answers by a user and returns it as string array
```
hanzei:
```suggestion
// getVotedAnswers collect voted answers by a user and returns it as string array.
```

server/poll/poll.go
```diff
@@ -190,8 +201,18 @@ func (p *Poll) Copy() *Poll {
 	return p2
 }
 
-// EncodeToByte returns a VotedAnswerResponse as a byte array
-func (v *VotedAnswerResponse) EncodeToByte() []byte {
+// EncodeToByte returns a Metadata as a byte array
```
hanzei:
```suggestion
// EncodeToByte returns a Metadata as a byte array.
```

server/poll/poll.go
```diff
@@ -190,8 +201,18 @@ func (p *Poll) Copy() *Poll {
...
+func (v *Metadata) EncodeToByte() []byte {
```
hanzei: Maybe we make this a `WriteTo(io.Writer) error` instead?

server/poll/transform.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/nicksnyder/go-i18n/v2/i18n"
 )
 
+const (
+	// MatterpollAdminButtonType is action_type of buttons that are used for managing a poll
```
hanzei:
```suggestion
	// MatterpollAdminButtonType is action_type of buttons that are used for managing a poll.
```

webapp/src/utils/constants.js
```diff
@@ -0,0 +1,5 @@
+export const ActionButtonType = {
+    BUTTON: 'button',
+    SELECT: 'select',
+    MATTERPOLL_ADMIN_BUTTON: 'custom_matterpoll_admin_button',
+};
```
hanzei: Nit: Add linebreak

server/plugin/api.go
kaakaa: Good point!

server/plugin/api.go
```diff
@@ -219,15 +220,7 @@ func (p *MatterpollPlugin) handleVote(vars map[string]string, request *model.Pos
...
-	v, err := poll.GetVotedAnswer(userID)
-	if err != nil {
-		return commandErrorGeneric, nil, errors.Wrap(err, "failed to get voted answers")
-	}
-	p.API.PublishWebSocketEvent("has_voted", map[string]interface{}{
-		"user_id":       v.UserID,
-		"poll_id":       v.PollID,
-		"voted_answers": v.VotedAnswers,
-	}, &model.WebsocketBroadcast{UserId: userID})
+	p.publishPollMetadata(poll, userID)
```
hanzei: Given that this takes some time, it might even be wroth to create a go routine for it. 0/5

server/poll/poll.go
```diff
@@ -32,11 +33,17 @@ type Settings struct {
...
-	UserID       string   `json:"user_id"`
-	VotedAnswers []string `json:"voted_answers"`
+// Metadata stores personaized metadata of a poll.
+//
+// PollID is ID for poll
+// UserID is ID for user
+// AdminPermission will be true if the user with "UserID" has admin permission for the poll with "PollID". Otherwise AdminPermission will be false.
+// VotedAnswers is list of answer that the user with "UserID" have voted for the poll with "PollID"
+type Metadata struct {
+	PollID          string   `json:"poll_id"`
```
hanzei: Instead of commenting before the struct, you can comment on the individual fields directly:
```suggestion
	PollID          string   `json:"poll_id"` // PollID is ID for poll
```

server/poll/poll.go
```diff
@@ -32,11 +33,17 @@ type Settings struct {
...
 
-// VotedAnswerResponse stores answers that is created by a user
-type VotedAnswerResponse struct {
-	PollID       string   `json:"poll_id"`
-	UserID       string   `json:"user_id"`
-	VotedAnswers []string `json:"voted_answers"`
+// Metadata stores personaized metadata of a poll.
+//
+// PollID is ID for poll
+// UserID is ID for user
```
hanzei: which user?

server/poll/poll.go
```diff
@@ -32,11 +33,17 @@ type Settings struct {
...
 }
 
-// VotedAnswerResponse stores answers that is created by a user
-type VotedAnswerResponse struct {
-	PollID       string   `json:"poll_id"`
-	UserID       string   `json:"user_id"`
-	VotedAnswers []string `json:"voted_answers"`
+// Metadata stores personaized metadata of a poll.
+//
+// PollID is ID for poll
```
hanzei: I don't think this needs to be documented.

server/poll/poll.go
```diff
@@ -32,11 +33,17 @@ type Settings struct {
...
+// UserID is ID for user
+// AdminPermission will be true if the user with "UserID" has admin permission for the poll with "PollID". Otherwise AdminPermission will be false.
+// VotedAnswers is list of answer that the user with "UserID" have voted for the poll with "PollID"
```
hanzei: :100:

server/poll/poll.go
```diff
@@ -190,8 +207,19 @@ func (p *Poll) Copy() *Poll {
...
 
-// EncodeToByte returns a VotedAnswerResponse as a byte array
-func (v *VotedAnswerResponse) EncodeToByte() []byte {
+// WriteTo write Metadata to the io.Writer given as an argument
+func (v *Metadata) WriteTo(w io.Writer) (int64, error) {
 	b, _ := json.Marshal(v)
-	return b
+	i, err := w.Write(b)
+	return int64(i), err
+}
```
hanzei: You can use [`json.Encoder`](https://golang.org/pkg/encoding/json/#NewEncoder) instead:
```suggestion
// WriteTo write Metadata to the io.Writer given as an argument
func (m *Metadata) WriteTo(w io.Writer) error {
	return json.NewEncoder(w).Encode(v)
}
```

server/poll/poll.go
```diff
@@ -190,8 +207,19 @@ func (p *Poll) Copy() *Poll {
...
+
+// ToMap returns a Metadata as a map
+func (v *Metadata) ToMap() map[string]interface{} {
```
hanzei: Nit:
```suggestion
func (m *Metadata) ToMap() map[string]interface{} {
```

server/poll/poll.go
```diff
@@ -32,11 +33,17 @@ type Settings struct {
...
-	PollID       string   `json:"poll_id"`
-	UserID       string   `json:"user_id"`
-	VotedAnswers []string `json:"voted_answers"`
+// Metadata stores personaized metadata of a poll.
+//
+// PollID is ID for poll
+// UserID is ID for user
+// AdminPermission will be true if the user with "UserID" has admin permission for the poll with "PollID". Otherwise AdminPermission will be false.
+// VotedAnswers is list of answer that the user with "UserID" have voted for the poll with "PollID"
+type Metadata struct {
```
hanzei: 0/5 maybe move this and the other related code to a new file `server/poll/metadata.go`

server/poll/poll.go
kaakaa: It looks smart way. But govet check if method `WriteTo(io.Writer)` have the signature like `WriteTo(io.Writer) (int64, error)`.

```
/usr/local/bin/go vet ./...
# github.com/matterpoll/matterpoll/server/poll
server/poll/poll.go:206:20: method WriteTo(w io.Writer) error should have signature WriteTo(io.Writer) (int64, error)
```

`json.NewEncoder(w).Encode(v)` returns only `error` object as return type, so it can't satisfy the rule for signature. We would have some way to fix this.

1. Change method name to avoid govet check (e.g.: `WriterToWriter`)
2. Adopt the existing implementation
3. Revert to `EncodeToByte`, and remove `WriteTo`

I think that **no.3 is the best** for us, because we don't need the size of data to be written. How does this sounds?

server/plugin/api.go
kaakaa: Good suggestion 👍

server/poll/poll.go
hanzei: To be fair, my comment was very nit picky in the first place. So, I'm 0/5 on this.

What about dropping this method and just use `json.Encoder` inside `api.go` in https://github.com/matterpoll/matterpoll/pull/234/files#diff-4444f0caa92e553b5c018c908e90e5e6R528?

hanzei:
Awesome work @kaakaa

PR #244, opened 2019-11-14
Tweak golangci config

hanzei:
Adds some useful linters and hardens the config. I also removed `gocyclo` because it's not a good metric.

Requires #243

### DIFF
--- a/server/plugin/plugin_test.go
+++ b/server/plugin/plugin_test.go
@@ -21,7 +21,7 @@ import (
 	"golang.org/x/text/language"
 )
 
-func setupTestPlugin(_ *testing.T, api plugin.API, store store.Store) *MatterpollPlugin {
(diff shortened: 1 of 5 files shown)


server/plugin/plugin_test.go
```diff
@@ -21,7 +21,7 @@ import (
 	"golang.org/x/text/language"
 )
 
-func setupTestPlugin(t *testing.T, api *plugintest.API, store *mockstore.Store) *MatterpollPlugin {
+func setupTestPlugin(_ *testing.T, api plugin.API, store store.Store) *MatterpollPlugin {
```
kaakaa: Why `plugintest.API` and `mockstore.Store` are replaced with `plugin.API` and  `store.Store`?

hanzei: Because `interfacer` suggested this :man_shrugging:

You are right, this is confusing. I will revert.

PR #246, opened 2019-11-14
Prevent spoofing of requests

### LINKED ISSUE #245: Requests to API can be spoofed if Experimental UI is enabled
If Experimental UI is enabled, we send a websocket event containing among other things the `pollID`. Any user receiving this websocket event can then use the `pollID` to handcraft events to the API and trigger events. E.g. he/she can vote a other users.

The easy solution is to include a `token` in `context` and check it on reviving. But this would be a breaking change.

I will be hard to fix this without introducing a breaking change. This can partly be fixed by checking if `Mattermost-User-ID` is the same as the `userID` in the request. With this change the user can only forge requests as himself. This still allows voting for channels the user is not part of any longer.

hanzei:
Fixes #245

### DIFF
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -144,6 +144,17 @@ func (p *MatterpollPlugin) handlePostActionIntegrationRequest(handler postAction
 			http.Error(w, "invalid request", http.StatusBadRequest)
 			return
 		}
+
+		if request.UserId != r.Header.Get("Mattermost-User-ID") {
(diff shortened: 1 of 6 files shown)


hanzei:
@kaakaa Please give this a final review

hanzei:
@kaakaa Please give this PR a review

kaakaa (approved):
Perfect 🎉  Thank you for finding this problem and elaborate in detail.

Issue #177, opened 2019-07-31
Add a ProgressBar Option to polls


tawalaya:
#### Summary

Hi, we use polls allot in our organization. One problem that we have with the otherwise excellent polls is that it is always difficult to see where a poll is currently going, visually. Additionally, even when a vote is over, it is hard to tell what the outcome is when scrolling by.

#### Describe alternatives you've considered

One way to improve this would be to add bars for each option. With this, it would be much easier to tell the outcome of a poll at a glance. 

![Mockup](https://imgur.com/pVIJE1j.png)

hanzei:
Hey @tawalaya,

Thanks for the suggestion. While I like the idea, it's hard to implement. Matterpoll is limited by the possibilities given my Mattermost. In the future we might be able to implement this via custom posts, but this will take time. See #160 for some details.

pykenny:
How about taking the approach used in [Polly](https://www.polly.ai/help/slack/creating-polls)? Polly uses inline code blocks filled with fullwidth squares to imitate progress bars, which I think that's relatively simpler to implement.

<img width="390" alt="image" src="https://github.com/matterpoll/matterpoll/assets/6328813/5901bc06-74bc-4a07-a59b-f0b6401c31b8"> ([Image Source](https://www.effy.ai/blog/slack-poll))

Few shortcomings I can think of right now:
1. Layout may be inconsistent amongst different devices/platforms (width/height of the bar, seam between characters, etc.)
2. Width of the bar should be carefully picked to avoid line break

Moreover, I guess the grid layout shown in the screenshot is implemented through [section fields](https://api.slack.com/reference/block-kit/blocks#section_fields). It has limit of holding at most 10 items per section block, so they'll probably stack multiple section blocks or use line-per-item as fallback when surpassing that limit (I've never tried stuffing a vote with > 10 options in Polly). Section fields also only allows placing text content within, so you can't place interactive components like buttons in the elements. Not certain if the grid layout is available in Mattermost's block layout, but line-per-item would should suffice in most cases.

hanzei:
Sure, that looks great @pykenny :+1:

PR #251, opened 2019-12-23
Added progress bars implementation of #177

tawalaya:
This implements a version of the feature requested in #177.
It's by no means perfect but I think it might already be beneficial to some. I tested it for desktop and web. It might look funny for mobile.

Screenshot_20191223_113752


server/poll/transform_test.go
```diff
@@ -97,7 +97,26 @@ func TestPollToEndPollPost(t *testing.T) {
...
 
+	for name, test := range map[string]struct {
+		Poll *poll.Poll
+	}{
+		"Test1": {
+			Poll: testutils.GetPollWithSettings(poll.Settings{Progress: true}),
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			test.Poll.UpdateVote(testutils.GetBotUserID(), 1)
```
golangcibot: Error return value of `test.Poll.UpdateVote` is not checked (from `errcheck`)

server/poll/transform_test.go
```diff
@@ -97,7 +97,26 @@ func TestPollToEndPollPost(t *testing.T) {
...
+			test.Poll.UpdateVote("bar", 1)
```
golangcibot: Error return value of `test.Poll.UpdateVote` is not checked (from `errcheck`)

server/poll/transform_test.go
```diff
@@ -97,7 +97,26 @@ func TestPollToEndPollPost(t *testing.T) {
...
+			test.Poll.UpdateVote("foo", 0)
```
golangcibot: Error return value of `test.Poll.UpdateVote` is not checked (from `errcheck`)

hanzei:
Thank you very much for the PR @tawalaya! I will review it after the holidays.

server/poll/transform.go
```diff
@@ -127,6 +136,44 @@ func (p *Poll) makeAdditionalText(localizer *i18n.Localizer, numberOfVotes int)
...
+	for _, n := range keys {
+		v := votes[n]
+		var progress float64
+		if numberOfVotes > 0 {
+			progress = float64(v) / float64(numberOfVotes)
+		} else {
+			progress = 0
+		}
+
+		lines = append(lines, fmt.Sprintf("%s\t%s\t`%3d %%`", progressBarStr(progress, 32), n, int(progress*100.0)))
```
kaakaa:
```suggestion
		lines = append(lines, fmt.Sprintf("`%s`\t%s\t`%3d %%`", progressBarStr(progress, 32), n, int(progress*100.0)))
```

In my browser, `█` and `░` seem to have different width. My suggestion would fix this difference, and additionally could fix issues on mobile. 
スクリーンショット 2020-01-07 22 26 47

server/poll/transform.go
```diff
@@ -127,6 +136,44 @@ func (p *Poll) makeAdditionalText(localizer *i18n.Localizer, numberOfVotes int)
...
+
+func generateProgressBars(votes map[string]int, numberOfVotes int) []string {
+	lines := make([]string, 0)
+
+	keys := make([]string, 0)
+	for n := range votes {
+		keys = append(keys, n)
+	}
+
+	sort.Strings(keys)
```
kaakaa: Is this line needed? This seems to be the cause of the wrong order between progress bars and answer buttons.

server/poll/transform.go
```diff
@@ -113,6 +117,11 @@ func (p *Poll) makeAdditionalText(localizer *i18n.Localizer, numberOfVotes int)
 	}
 
 	lines := []string{"---"}
+
+	if p.Settings.Progress {
+		lines = append(lines, generateProgressBars(votes, numberOfVotes)...)
```
kaakaa: I prefer this line below "TotalVotes" line. [0/5](https://docs.mattermost.com/process/training.html#list-of-terms)

server/poll/transform.go
tawalaya: What does the ``` do exactly? Good point about the width of these characters, it worked for the clients I tested it with. There are a few different symbols like this but, I think this is a font issue that will make it difficult to select characters that work the same in all fonts. Any ideas?

server/poll/transform.go
tawalaya: I was unsure if the order was stable or not for this parameter and thus did a sort to ensure that it always stays in the same order. If its always the same order as the buttons and it remains stable than it can be removed for sure.

server/poll/transform.go
kaakaa: I'm sorry I don't have any better ideas...

Surrounding words with `` `TEXT` `` is the format of Markdown, and is converted to `<span class="codespan__pre-wrap"><code>TEXT</code></span>` by Mattermost. In my local, although normal text is rendered with Open Sans font, `<code>` elements is rendered with monospace fonts (e.g. Menlo, Monaco, monospace, or so on). But as you pointed out, which font is actually used would depend on the user's environment. So my idea is just a workaround, but I think it would be some effective with less efforts.

server/poll/transform.go
kaakaa: You're right. The order of keys from map is unstable, so without sorting keys the order of progress bar will change with each run... However, buttons is not ordered alphabetically, so it's going to be particularly weird when adding answers.

So what do you think about for-looping `poll.AnswerOption` here instead of creating `votes` in `ToPostActions`?  `poll.AnswerOption` is slice and the order of elements in slice would be preserved, so without the use of map the order of buttons and progress bars would be matched.

I know it is duplicated for-loop for `poll.AnswerOption`, but it would not be a performance issue because the number of answer options would be a few in most cases.

server/poll/transform_test.go
```diff
@@ -97,7 +98,53 @@ func TestPollToEndPollPost(t *testing.T) {
...
+			lines := strings.Split(postText, "\n")
+			require.GreaterOrEqual(t, len(lines), 4)
+
+			filled := strings.Count(lines[1], "█")
+
+			filled += strings.Count(lines[2], "█")
+
+			filled += strings.Count(lines[3], "█")
+
+			//This value should be close to the total length of a progress bar (32 chars), it might be a littel less due to rounding errors
```
golangcibot: `littel` is a misspelling of `little` (from `misspell`)
```suggestion

```

tawalaya:
As a quick summary, I added the changes @kaakaa suggested. Also, I extended the test for this functionality as it did not do much before :sweat_smile:.

kaakaa:
Thank you very much @tawalaya!  It looks good 👍 

@hanzei Could you please review this?

hanzei:
@kaakaa I will review the PR over the weekend

tawalaya:
@hanzei No but it's the same for me on some devices. I think it depends on some font scaling thing thats outside of the control of mattermost. 

On mobile it still looks kind of bad, see image below.
However, there is prob. no good way to solve this issue. We can reduce the number of blocks but depending on the screen size of the device it will look like that for some devices. 

Any good ideas? Is there a way to detect if this is rendered on mobile?

Screenshot_20200215-162920_Mattermost

kaakaa:
Thanks @hanzei and @tawalaya .
I investigated approaches to use PNG and SVG for mobile app. 

Let me get straight to the point, PNG approach is not applicable. SVG approach is also not applicable by so far, but **after merging [WIP: MM\-9746 \- display markdown svgs](https://github.com/mattermost/mattermost-mobile/pull/3858), SVG approach might be applicable**.

### PNG approach

PNG approach is to use a thin image file and specify their width in inline markdown image like the following.

```
![bar](/plugins/com.github.matterpoll.matterpoll/bar.png =450x10)
```

[M-Mueller/mattermost-poll](https://github.com/M-Mueller/mattermost-poll) uses this approach [here](https://github.com/M-Mueller/mattermost-poll/blob/f54b8200e1e97855cf8ab8f00c54cd6a34290c0e/formatters.py#L157).

It can be rendered in iOS, chrome and safari. **However, this markdown format to specify the size of image at the end does not seem to be available in mobile apps.**

IMG_5037

So I supposed that PNG approach is not applicable.

### SVG approach

SVG approach is to use a .svg file like the following (, and it had better to use a template file instead of a svg file in order to specify the width of rect in runtime).

```svg
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewbox="0 0 1 1" x="0" y="0" width="100%" height="10">
    <rect fill='grey' x="0" y="0" width='100%' height="10" />
    <rect fill='blue' x="0" y="0" width='50%' height="10" />
</svg>
```

It can be rendered in iOS, chrome and safari.

IMG_65BD5F219920-1

**However, inline markdown svg cannot be rendered in mobile app, because PR for inline markdown svg is now `open` in https://github.com/mattermost/mattermost-mobile/pull/3858 .**

IMG_5038

---

So I supposed that it had better to try to implement this feature after merging [WIP: MM\-9746 \- display markdown svgs](https://github.com/mattermost/mattermost-mobile/pull/3858). How does this sounds?

hanzei:
Thanks for the investigation @kaakaa. :+1: for the svg approach.

I've found https://github.com/wcharczuk/go-chart which might be a good library to generate the svg charts.

tawalaya:
Hello,
I had some time on my hands and, based on @hanzei idea, implemented something with images instead of ASCII. Since SVGs don't work, I just generated PNGs for now, not ideal, but as a proof-of-concept, they work on both mobile and desktop.

At the moment and size, colour is hardcoded also. There is no caching of the generated images, though it's not that heavy. 
What do you think? Any guidance on how to use theme information and also how to proceed would be welcome.

On Web/Desktop:
image

On Android:
Screenshot_20211010-223901

server/plugin/api.go
```diff
@@ -72,11 +72,12 @@ func (p *MatterpollPlugin) InitAPI() *mux.Router {
 	r := mux.NewRouter()
 	r.HandleFunc("/", p.handleInfo).Methods(http.MethodGet)
 	r.HandleFunc("/"+iconFilename, p.handleLogo).Methods(http.MethodGet)
+	// INFO we need to it here, because we would have to put an exception to checkAuthenticity otherwiese
```
kaakaa, line 75: I tested in Mattermost v6, and I found that the request from message attachment to server seems to send with auth info. So I think this endpoint can/should be placed under `/api/v1`.

kaakaa (commented):
Thanks for restarting this again. It looks Interesting.

I don't know how to use theme color on server, but I feel like the frame makes it recognizable.
スクリーンショット 2021-11-03 12 25 50

The size of a bar image is a few hundred bytes at most, so it would be nice to be able to cache images, but I don't think it's a must. 

What do you think about this? @hanzei

kaakaa:
Another problem: iOS mobile app doesn't display images, even though iOS chrome do it. I'll take a look into this.
image

hanzei:
@tawalaya Do you want to continue working on the PR?

tawalaya:
> @tawalaya Do you want to continue working on the PR?

I didn't pressure this further as it was unclear if we wanted the ASCII method (needs a decision maximum width) or the issue of styling the image correctly. What is the best approach to move forward? 

If you have an idea, let me know, and I can spend some time to refresh the PR.

tawalaya:
Hi there (@kaakaa, @hanzei), I've reworked the PR as suggested, focusing on the ASCII part.  Thanks to @mickmister for helping with fixing an issue with the web app part of the plugin! I can remove that part again before the final version of this PR. 

To address the issue with the rendering size, I think the best possible solution is to let the server admin decide. So, I added some settings for it. 

This is what it looks like now on Desktop:
Screenshot 2024-01-12 at 17 06 20

and Mobile:
Screenshot_20240112-170644


I've also explored the option proposed by @pykenny
Desktop: 
Screenshot 2024-01-12 at 17 03 18
Mobile:
Screenshot_20240112-170308

Now, the issue with setting the right length will remain. Let me know what you think.

kaakaa:
Thanks @tawalaya! ASCII way is work fine for me, and making the rendering size configurable looks a great idea. In terms of UI, I prefer the first example without the use of AttachmentField, because bars in a row are easier to see the differences and (as previously mentioned) bars are collapsed when there are a large number of answer options.

@hanzei how do you think about this way?

I also leave a note of my opinion for current implementation.

* When using the multi votes setting (`--votes=X`), progressbars doesn't seem to work correctly. To calculate bar, `numberOfVoters` should be used instead of `numberOfVotes`, which being introduced in #517.
* I think that progressbar should be displayed below the `Poll settings` and `Total votes`.
* I wonder if introducing the new `--progress-bar` setting with only the functionality of showing progressbar added to `--progress` setting instead of introducting ShowProgressBar config. Since showing progressbars needs some spaces, it may be better if the creator of the poll could choose.

kaakaa:
@hanzei Gentle reminder, what's your take on this?

kaakaa:
@tawalaya Sorry for the frequent late replies. I would like to move this PR forward, but how about [points I commented before](https://github.com/matterpoll/matterpoll/pull/251#issuecomment-1890966310)? I would like to start a code review after that points is resolved.

tawalaya:
@kaakaa I've looked into it. I've already addressed the first two points. The last one  (`--progress-bar`) started to work on, but since it affects the command options, it has a large tail of changes in the testing code so I've not finished that one until it was clear if it should go into that direction. But I can spent some time to work on it in a few days.

hanzei:
> * I wonder if introducing the new `--progress-bar` setting with only the functionality of showing progressbar added to `--progress` setting instead of introducting ShowProgressBar config. Since showing progressbars needs some spaces, it may be better if the creator of the poll could choose.

I tend to disagree. Polls should look uniform on a server. Given that the ASCI art looks a bit off, I think it makes sense to have an system-level option to disable/enable the progress bar.

kaakaa:
Thank you for your comments @tawalaya  @hanzei.  Okay, I drop the idea of introducing the `--progress-bar` and let's keep the `ShowProgressBar` in the system settings. Sorry for the confusion, but has the direction become clear? @tawalaya

tawalaya:
Hi there, I've implemented the suggestions. Have a look. The current test is a bit "hacky" tbh. feel free to improve it using the mocking features used in the other tests.

webapp/tests/setup.tsx
```diff
@@ -1,7 +1,11 @@
-import 'mattermost-webapp/tests/setup';
 import '@babel/polyfill';
 import {jest} from '@jest/globals';
 
+import Enzyme from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
```
hanzei, line 5: CI is not happy

```
Error: tests/setup.tsx(5,21): error TS7016: Could not find a declaration file for module 'enzyme-adapter-react-16'. '/home/runner/work/matterpoll/matterpoll/webapp/node_modules/enzyme-adapter-react-16/build/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/enzyme-adapter-react-16` if it exists or add a new declaration (.d.ts) file containing `declare module 'enzyme-adapter-react-16';`
make: *** [Makefile:49: check-style] Error 2
Error: Process completed with exit code 2.
```
https://github.com/matterpoll/matterpoll/actions/runs/8589443726/job/23559354710?pr=251

kaakaa (changes requested):
Thanks a lot @tawalaya! I've left some suggestions and comments, please confirm those.

server/poll/transform.go
```diff
@@ -192,9 +195,48 @@ func (p *Poll) makeAdditionalText(bundle *utils.Bundle, numberOfVotes, numberOfV
 			TemplateData:   map[string]interface{}{"TotalVotes": numberOfVotes},
 		}))
 	}
+
+	if p.Settings.Progress && p.Settings.ShowProgressBars {
+		lines = append(lines, "---")
```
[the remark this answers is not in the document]
tawalaya: Wow, good spot! yes.

server/poll/transform.go
```diff
@@ -192,9 +195,48 @@ func (p *Poll) makeAdditionalText(bundle *utils.Bundle, numberOfVotes, numberOfV
...
+	for i := 0.0; i < wholeWidth; i++ {
+		line += "█"
+	}
+	for i := 0.0; i < (width - wholeWidth); i++ {
+		line += "░"
+	}
+	return line
+}
+
+func generateProgressBars(answerOptions []*AnswerOption, numberOfVotes, MAX_BAR_CHAR_LENGTH int) []string {
```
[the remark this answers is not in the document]
tawalaya: Yep, that used to be a constant... will revise 👍

webapp/tests/setup.tsx
tawalaya: I didn't really touch JS other than to test it on Mac. I needed to pull in an update at some point to make it build. Could be related to that. Any advise on how to fix this? I think I'll just take revert to the current master for all files in webapp to be sure.

server/poll/transform.go
```diff
@@ -192,9 +195,48 @@ func (p *Poll) makeAdditionalText(bundle *utils.Bundle, numberOfVotes, numberOfV
 			TemplateData:   map[string]interface{}{"TotalVotes": numberOfVotes},
 		}))
 	}
+
+	if p.Settings.Progress && p.Settings.ShowProgressBars {
+		lines = append(lines, "---")
+		if p.IsMultiVote() {
+			lines = append(lines, generateProgressBars(p.AnswerOptions, numberOfVotes, p.Settings.ProgressBarLength)...)
```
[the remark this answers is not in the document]
tawalaya: But is that right? This way, we could have two options with 100. If User C also voted for option 2?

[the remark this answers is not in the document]
kaakaa: Is there something blocking this? Is the current implementation better?

Issue #255, opened 2020-02-07
Multiple choice?


Domvel:
A multiple choice option would be nice. One user should vote for multiple options. Currently an user can only update his vote. The multiple choice vote should toggle the option. If possible highlight the selected option with a background color or another border color or just add something in the text of the option to indicate a vote for the user. e.g. an asterisk. (The feature to highlight current vote can be done in another issue. But it's important especially for multiple choices. If this is not possible for technical reasons, we could find another solution. for example: Improve the poll messages after a vote with markdown stuff or colors if possible.)

Current options:

- `--anonymous`: Don't show who voted for what
- `--progress`: During the poll, show how many votes each answer option got
- `--public-add-option`: Allow all users to add additional options.

New:

- `--multiple-choice`: Allow multiple choice.

(Or just `--multiple` or `--m` ... maybe aliasses.)

kaakaa:
Thank you for commenting @Domvel . I start implementing multi answer option according to [this comment](https://github.com/matterpoll/matterpoll/issues/84#issuecomment-543342123).
My implementation will achieve multi choice by `--votes=X` option (`X` is the number of vote that user can give). And if you enable `Experimental UI` setting for Matterpoll, selected option will be highlighted.

But the toggle feature will not be implemented, because the feature to highlight option is now experimental feature, and we don't consider experimental features by default.  Additionally the reason why highlighting is experimental is that we don't have reviewers for webapp.

Domvel:
@kaakaa Thanks. I missed this issue #84 on Github search. 
You could use Checkmarks on the left or right of an answer option. If highlight is only available in `Experiemental UI`. But I don't know if the manipulating of the options is possible. Only for the voter and not for all. Or write the current state in the bot message "Your bote has been updated." with the current votes. Just a spontaneous idea.

kaakaa:
Thanks @Domvel!

If Checkmarks can be visible only for the voter, it must be handled by client-side (webapp plugin), and it will be also included `Experimental UI` settings. And I suppose that highlighting button is easier to see than checkmarks.

`ExperimentalUI` can be configured from **System Console > Plugins > Matterpoll**, and only system admin may be enable to access `System Console`.  Please try to use Experimental UI.

https://github.com/matterpoll/matterpoll#settings
> You can configure Matterpoll from System Console > Plugins > Matterpoll.
>   * Trigger Word: Change trigger word for poll command. (default /poll)
[...]

PR #256, opened 2020-02-11
Implement "--votes=X" setting for multi answer

### LINKED ISSUE #84: Multi answer option
Now, users can vote just one answer.
It had better to add settings that users can vote multi answers.

kaakaa:
Fixed #84 

* Add `--votes=X` setting for poll
  * Validate that `X` is a positive number or less than the number of answer options
* If `votes` setting is enabled, voting will return message according to [this comment](https://github.com/matterpoll/matterpoll/issues/84#issuecomment-543342123)
* If `votes` setting is enabled,`Reset Votes` button will be created to reset all existing votes
* Experimantal UI can highlight multi answers

### TODO
- ~~[ ] handle options that enable to vote all options (even if options are added)~~
- [x] update interactive dialog


server/plugin/api_test.go
```diff
@@ -382,6 +447,218 @@ func TestHandleVote(t *testing.T) {
...
+		AnswerOptions: []*poll.AnswerOption{
+			{Answer: "Answer 1", Voter: []string{}},
+			{Answer: "Answer 2", Voter: []string{}},
+			{Answer: "Answer 3", Voter: []string{}},
+		},
+		Settings: poll.Settings{MaxVotes: 3},
+	}
+
+	poll2WithVotes := poll.Copy()
+	poll2WithVotes.UpdateVote("userID1", 0)
```
golangcibot: Error return value of `poll2WithVotes.UpdateVote` is not checked (from `errcheck`)

server/plugin/api_test.go
```diff
@@ -382,6 +447,218 @@ func TestHandleVote(t *testing.T) {
...
+
+	poll3WithVotes := poll.Copy()
+	poll3WithVotes.UpdateVote("userID1", 0)
```
golangcibot: Error return value of `poll3WithVotes.UpdateVote` is not checked (from `errcheck`)

server/plugin/api_test.go
```diff
@@ -382,6 +447,218 @@ func TestHandleVote(t *testing.T) {
...
+	poll3WithVotes.UpdateVote("userID1", 1)
```
golangcibot: Error return value of `poll3WithVotes.UpdateVote` is not checked (from `errcheck`)

server/plugin/api_test.go
```diff
@@ -382,6 +447,218 @@ func TestHandleVote(t *testing.T) {
...
+	poll3WithVotes.UpdateVote("userID1", 2)
```
golangcibot: Error return value of `poll3WithVotes.UpdateVote` is not checked (from `errcheck`)

server/plugin/api_test.go
```diff
@@ -382,6 +447,218 @@ func TestHandleVote(t *testing.T) {
...
+
+	poll4WithVotes := poll.Copy()
+	poll4WithVotes.UpdateVote("userID1", 0)
```
golangcibot: Error return value of `poll4WithVotes.UpdateVote` is not checked (from `errcheck`)

kaakaa:
https://github.com/matterpoll/matterpoll/issues/255#issuecomment-586987176
> > For the option votes there should be a value to allow vote all. For example the value 0 or if omitted value. (Nice to have. Otherwise: Just set the value to my options or more. Default is 1 I guess?)
> 
> You pointed out good points. Now, with #256 users can specify the positive number between 1 and the number of options as --votes=X, but it can't handle options added after the poll was created. We'll need to discuss this in #256. Thanks.

kaakaa:
@hanzei 
I would like to suggest to introduce new `--vote-all` setting besides `--votes=X` for [this point](https://github.com/matterpoll/matterpoll/pull/256#issuecomment-587037167)
* if `--vote-all` is set, user can vote on **ALL** options, even if options are added
* if `--votes=X` is set, user can vote on **X** options at any time, even if options are added
* If the both of `--vote-all` and `--votes=X` are set, `--vote-all` takes precedence (or return error?)

What do you think about this?

server/poll/poll.go
```diff
@@ -30,6 +34,7 @@ type Settings struct {
 	Anonymous       bool
 	Progress        bool
 	PublicAddOption bool
+	MaxVotes        int
```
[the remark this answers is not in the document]
kaakaa: If the type of `MaxVotes` is changed to `uint`, `MaxVotes` must always be calculated / compared with `uint` value. For example, we must write like `MaxVotes <= uint(len(votedAnswers))`. It looks little ugly. Also, if the value of `MaxVotes` becomes smaller than 0, overflow immediately occurs. So I think that using `uint` here is not good for us.

hanzei:
@kaakaa How would a `--vote-all` setting be shown in the dialog from `/poll`? Having two setting in the dialog there could be confusing.

kaakaa:
@hanzei That might be true.  Current implementation shows the following dialog, and It may seem like too much of options.

スクリーンショット 2020-03-25 22 45 46

Which option looks better for all votes, `--vote-all` or `--votes=0`? or other?

kaakaa:
@hanzei 
As we've seen in [user comments](https://github.com/matterpoll/matterpoll/pull/256#issuecomment-587037167), I think it's important for users to be able to vote on every answer, even if options are added by `add option` feature. However, since there are so many changes and the release will be delayed, it might be better to turn it over to the next implementation.

So I would like to complete this PR with neither `-vote-all` nor `-votes=0`.

kaakaa:
@hanzei I supposed that this feature have been implemented now. If you have a time, could you give me feedbacks? (And I am aware that there is still a task to add a few more tests.)

server/poll/poll.go
```diff
@@ -75,6 +85,36 @@ func NewPoll(creator, question string, answerOptions, settings []string) (*Poll,
...
+				"Setting": s,
+			},
+		}
+	}
+	i, err := strconv.Atoi(e[1])
+	if err != nil || i <= 0 || i > len(p.AnswerOptions) {
+		return &ErrorMessage{
+			Message: &i18n.Message{
+				ID:    "poll.newPoll.votesettings.invalidSetting",
+				Other: "In votes=X, X must be a positive number and less than the number of answer options. {{.Setting}}",
```
hanzei, line 107: That error message looks weird in the modal. Could it me more generic and the command handlers adds it's specific information itself?
Screenshot from 2020-05-29 10-56-56

server/poll/poll.go
```diff
@@ -75,6 +85,36 @@ func NewPoll(creator, question string, answerOptions, settings []string) (*Poll,
 	return &p, nil
 }
 
+// ParseVotesSetting parses and sets a setting for votes ("--votes=X")
+func (p *Poll) ParseVotesSetting(s string) *ErrorMessage {
+	e := votesSettingPattern.FindStringSubmatch(s)
+	if len(e) != 2 {
```
hanzei, line 91: I would really like to see a test for this case.

server/poll/poll.go
```diff
@@ -108,26 +148,69 @@ func (p *Poll) AddAnswerOption(newAnswerOption string) *ErrorMessage {
...
+			}
+		}
 	}
+
+	p.AnswerOptions[index].Voter = append(p.AnswerOptions[index].Voter, userID)
+	return nil, nil
+}
+
+// ResetVotes remove votes by a given user
+func (p *Poll) ResetVotes(userID string) error {
```
hanzei, line 196: Similar comment as for `UpdateVote`: Is `error` really valuable here?

server/poll/transform.go
```diff
@@ -75,6 +79,18 @@ func (p *Poll) ToPostActions(localizer *i18n.Localizer, pluginID, authorName str
 		})
 	}
 
+	if p.Settings.MaxVotes > 1 {
+		actions = append(actions,
+			&model.PostAction{
+				Id:   "resetVote",
+				Name: localizer.MustLocalize(&i18n.LocalizeConfig{DefaultMessage: pollButtonResetVotes}),
```
hanzei, line 86: Similar comment as above: Maybe inline the `i18n.Message`here

server/plugin/api.go
```diff
@@ -305,10 +329,22 @@ func (p *MatterpollPlugin) handleVote(vars map[string]string, request *model.Pos
 	model.ParseSlackAttachment(post, poll.ToPostActions(publicLocalizer, manifest.ID, displayName))
 	post.AddProp("poll_id", poll.ID)
 
+	if poll.Settings.MaxVotes > 1 {
```
hanzei, line 332: Nit pick: Maybe add `poll.IsMultiVote() bool` as a helper.

server/plugin/api.go
```diff
@@ -326,22 +362,66 @@ func (p *MatterpollPlugin) publishPollMetadata(poll *poll.Poll, userID string) {
...
+	if err != nil {
+		return &i18n.LocalizeConfig{DefaultMessage: commandErrorGeneric}, nil, errors.Wrap(err, "failed to get voted answers")
+	}
+	if len(votedAnswers) == 0 {
+		return &i18n.LocalizeConfig{DefaultMessage: responseResetVotesNoVotes}, nil, nil
+	}
+
+	prev := poll.Copy()
+	if err = poll.ResetVotes(userID); err != nil {
+		return &i18n.LocalizeConfig{DefaultMessage: commandErrorGeneric}, nil, err
```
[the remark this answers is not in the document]
kaakaa: careless...

server/plugin/command.go
```diff
@@ -227,6 +232,15 @@ func (p *MatterpollPlugin) getCreatePollDialog(siteURL, rootID string, l *i18n.L
 			Optional: i > 2,
 		})
 	}
+
+	elements = append(elements, model.DialogElement{
+		DisplayName: "Multi vote",
+		Name:        "setting-multi",
+		Type:        "text",
+		SubType:     "number",
+		HelpText:    "The number of options that an user can vote on. If empty, users are allowed to vote only one option.",
```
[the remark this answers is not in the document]
kaakaa: It makes so clear 👍

server/plugin/api.go
```diff
@@ -228,7 +243,12 @@ func (p *MatterpollPlugin) handleCreatePoll(_ map[string]string, request *model.
 
 	var settings []string
 	for k, v := range request.Submission {
-		if strings.HasPrefix(k, "setting-") {
+		if k == "setting-multi" {
+			f, ok := v.(float64)
+			if ok {
+				settings = append(settings, fmt.Sprintf("votes=%d", int(f)))
```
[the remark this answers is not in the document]
kaakaa: I agree with you, but I also don't have an idea. I'll think about it again at a later.

server/plugin/api.go
```diff
@@ -42,6 +42,20 @@ var (
 		Other: "Your vote has been updated.",
 	}
 
+	responseVoteUpdatedMulti = &i18n.Message{
+		ID:    "response.vote.multi.updated",
+		One:   "Your vote has been counted. You have {{.Remains}} vote left.",
+		Other: "Your vote has been counted. You have {{.Remains}} votes left.",
```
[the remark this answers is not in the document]
kaakaa: Should I add `Few` and `Many` here? If so, I should add `Few` / `Many` in `poll.endPost.answer.heading`, right?

server/poll/poll.go
kaakaa: X is a temporary number, and the value is not embedded.
It's certainly weird to see it in dialog. I'm going to amend it to say, `The number of votes must be a positive number and less than or equal to the number of options. You specified "votes=50."`

kaakaa:
Thank you for great feedbacks 👍  I've fixed some of the points, but I'll address the rest at a later date.

server/poll/poll.go
kaakaa: It doesn't look valuable, so remove it.

But as for `UpdateVote`, it should check if  optionNumber is valid, because there may be the potential for inconsistency in the matterpoll and the potential for terrible errors if the value is incorrect. Also, if don't check the user ID, the user ID as empty will be added as a voter. (I was able to notice this by tests)

server/plugin/api.go
```diff
@@ -305,10 +329,22 @@ func (p *MatterpollPlugin) handleVote(vars map[string]string, request *model.Pos
 	model.ParseSlackAttachment(post, poll.ToPostActions(publicLocalizer, manifest.ID, displayName))
 	post.AddProp("poll_id", poll.ID)
 
+	if poll.Settings.MaxVotes > 1 {
+		// Multi Answer Mode
+		votedAnswers, _ := poll.GetVotedAnswers(userID)
```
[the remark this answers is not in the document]
kaakaa: nice 👍

server/poll/poll.go
```diff
@@ -108,26 +148,69 @@ func (p *Poll) AddAnswerOption(newAnswerOption string) *ErrorMessage {
 }
 
 // UpdateVote performs a vote for a given user
-func (p *Poll) UpdateVote(userID string, index int) error {
+func (p *Poll) UpdateVote(userID string, index int) (*i18n.Message, error) {
```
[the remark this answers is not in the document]
kaakaa: Sounds good. However, [caller](https://github.com/matterpoll/matterpoll/pull/256/files#diff-4444f0caa92e553b5c018c908e90e5e6R298) expects i18n.Message to be returned, so I don't think it's necessary to wrap it with `ErrorMessage`.  1/5

hanzei:
@kaakaa Heads up that there is a merge conflict to resolve

kaakaa:
@hanzei Thanks! Can you give this a second round of review?

kaakaa:
I found that this PR introduce breaking changes that existing polls never be updated. 
After this change, when voting to existing polls, it return an error that is generated here.
https://github.com/matterpoll/matterpoll/blob/master/server/store/kvstore/poll.go#L64

I supposed that the cause may be to add `MaxVotes` field to `poll.Settings`. As I need to investigate the behavior of `KVSetWithOptions`, I labeled this WIP.

ArpitKotecha:
@hanzei @kaakaa Any update on this? Looking forward to this feautre.

server/plugin/api.go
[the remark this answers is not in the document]
hanzei: Added in 3788dda

hanzei:
@kaakaa Yes, you are right a DB migrating was needed. I went ahead an implemented it myself so we can get this PR merged soon. Please take a look at [`e7cbc45` (#256)](https://github.com/matterpoll/matterpoll/pull/256/commits/e7cbc45aef9e488fef5f359704f0c8e5949a4cc9).

I will also do a final review myself.

server/plugin/api.go
```diff
@@ -317,31 +331,72 @@ func (p *MatterpollPlugin) publishPollMetadata(poll *poll.Poll, userID string) {
...
+
+	post := &model.Post{}
+	publicLocalizer := p.getServerLocalizer()
+	model.ParseSlackAttachment(post, poll.ToPostActions(publicLocalizer, manifest.ID, displayName))
+	post.AddProp("poll_id", poll.ID)
+
+	return &i18n.LocalizeConfig{
+		DefaultMessage: &i18n.Message{
+			ID:    "response.resetVotes.success",
+			Other: "All votes are cleared. [{{.ClearedVotes}}]",
```
hanzei, line 378: I'm wondering what the value of printing the previous answers is. The user already reset them, why do we print them?

hanzei (approved):
Awesome, let's get this merged. :rocket: 

One question below

server/store/kvstore/upgrade.go
```diff
@@ -62,3 +71,48 @@ func (s *Store) shouldPerformUpgrade(currentSchemaVersion, expectedSchemaVersion
...
+func upgradeTo14(s *Store) error {
+	var allKeys []string
+	i := 0
+	for {
+		keys, appErr := s.api.KVList(i, perPage)
+		if appErr != nil {
+			return errors.Wrap(appErr, "failed to list poll keys")
+		}
+
+		if len(keys) == 0 {
```
kaakaa, line 84:
```suggestion
		if len(keys) < perPage {
```

This would reduce the number of invoking `KVList` by one in many cases

server/store/kvstore/upgrade_test.go
```diff
@@ -140,3 +143,74 @@ func TestStoreUpdateDatabase(t *testing.T) {
 		assert.NotNil(t, err)
 	})
 }
+
+func TestUpgradeTo14(t *testing.T) {
+	t.Run("KVList succeeds", func(t *testing.T) {
```
kaakaa, line 149: Good 👍

kaakaa (commented):
Great 👍  Thanks @hanzei.  Left a tiny suggestion.

server/plugin/api.go
kaakaa: That isn't necessary, but I imagined that there would be cases where voter would want to see where voter had voted after resetting, such as if voter wanted to change only one of the many voted options. Should remove it?

hanzei: I'm 2/5 that the current text isn't great. If we keep it, can we change it to something like `All votes are cleared. Your previous votes were {{.ClearedVotes}}.`?

server/store/kvstore/upgrade.go
hanzei: Good point :+1: Updated

server/plugin/api.go
kaakaa: It looks so good! I've fixed as you suggested. Thanks.

hanzei:
Finally :rocket:

arashgmn:
> @hanzei As we've seen in [user comments](https://github.com/matterpoll/matterpoll/pull/256#issuecomment-587037167), I think it's important for users to be able to vote on every answer, even if options are added by `add option` feature. However, since there are so many changes and the release will be delayed, it might be better to turn it over to the next implementation.
> 
> So I would like to complete this PR with neither `-vote-all` nor `-votes=0`.

Sorry to open this again. But I couldn't figure out how can I configure my poll to allow users to vote possibly to all (even the newly added) options. To be clear, `--votes=X` works as expected. I'm thinking of the case when the total number of options is greater than X.

hanzei:
@arashgmn That use case is currently not supported. Please open a new feature request.

Issue #252, opened 2020-01-10
Cannot close poll when default language of server is polish


amz:
#### Summary
Cannot close poll

#### Steps to reproduce
1. Cannot close poll
2. Error ocurres

#### Expected behavior
Can close poll

#### Observed behavior (that appears unintentional)
After confirming closing in popup http request is made to:
/api/v4/actions/dialogs/submit with emtpy {} response and http status 200

In admin console -> plugin management i can see:
Matterpoll (com.github.matterpoll.matterpoll - 1.3.0)
This plugin crashed multiple times and is no longer running. Check your system logs for errors.

#### Server configuration
**Operating system:**
Ubuntu 18.04.3
**Mattermost version:**
5.18.1
**Matterpoll version:**
1.3.0
**Updated from an older Matterpoll version or fresh install:**
Updated from 5.14. At 5.14 was similar problem
<details>
<summary>Mattermost server logs</summary>

```
Jan 10 20:10:54 localhost mattermost[35696]: {"level":"debug","ts":1578687054.2560847,"caller":"mlog/sugar.go:15","msg":"New request:","plugin_id":"com.github.matterpoll.matterpoll","Host":"","RequestURI":"","Method":"POST"}
Jan 10 20:10:54 localhost mattermost[35696]: {"level":"debug","ts":1578687054.7189283,"caller":"web/handlers.go:77","msg":"Received HTTP request","method":"POST","url":"/api/v4/users/status/ids","request_id":"cnhorea46t8kpkb9dn53hifa3r"}
Jan 10 20:10:55 localhost mattermost[35696]: {"level":"debug","ts":1578687055.574891,"caller":"web/handlers.go:77","msg":"Received HTTP request","method":"POST","url":"/api/v4/users/status/ids","request_id":"bzsm7kc6aj8zip1a9my3371m6w"}
Jan 10 20:10:56 localhost mattermost[35696]: {"level":"debug","ts":1578687056.679782,"caller":"web/handlers.go:77","msg":"Received HTTP request","method":"POST","url":"/api/v4/actions/dialogs/submit","request_id":"xbd83y3xjpbrzfy6r9anwu9yya"}
Jan 10 20:10:56 localhost mattermost[35696]: {"level":"debug","ts":1578687056.6808593,"caller":"mlog/sugar.go:15","msg":"New request:","plugin_id":"com.github.matterpoll.matterpoll","Host":"","RequestURI":"","Method":"POST"}
[...]

PR #259, opened 2020-02-28
Add missing translations for polish

hanzei:
Add `few` and `many` translation for polish.

@amz @burasuk I don't know polish and can't confirm that this is a valid translation. Would you please give this a quick look?

Fixes https://github.com/matterpoll/matterpoll/issues/252

### DIFF
--- a/assets/i18n/active.pl.json
+++ b/assets/i18n/active.pl.json
@@ -82,9 +82,9 @@
     "poll.endPost.answer.heading": {
         "hash": "sha1-46743e9dbbeec0c43d16fe387735cf95c2e27abb",
         "one": "{{.Answer}} ({{.Count}} głos)",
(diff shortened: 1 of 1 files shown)


amz:
Yes, I know polish. But newer have translated using cldr rules. Based on http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html
I  would consider:

 "few": "{{.Answer}} ({{.Count}} głosy)",
 "many": "{{.Answer}} ({{.Count}} głosów)",
"other": "{{.Answer}} ({{.Count}} głosu)"

I can prepare a pull request if You wish.

hanzei:
@amz Thanks for the help. I've updated the translations. 

The polish translator @burasuk doesn't seam to be to active. Would you be interested in helping with the polish translations?

Issue #262, opened 2020-02-28
Make use of EnsureBotOption


hanzei:
See https://github.com/mattermost/mattermost-server/pull/12153

PR #263, opened 2020-02-28
Make use of EnsureBotOption

hanzei:
Fixes https://github.com/matterpoll/matterpoll/issues/262

### DIFF
--- a/server/plugin/plugin.go
+++ b/server/plugin/plugin.go
@@ -2,8 +2,6 @@ package plugin
 
 import (
 	"fmt"
-	"io/ioutil"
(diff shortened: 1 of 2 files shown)


hanzei:
Unfortunately, it is. Before bot accounts where used,  [/plugins/com.github.matterpoll.matterpoll/logo_dark.png](https://github.com/matterpoll/matterpoll/blob/c2ddb7877760f9823ac7b3694ea9cf8ddd12f76a/server/plugin/api.go#L73) was used as the path to the overwritten user icon. In order to not break existing polls, we have to keep this API method forever.

Issue #269, opened 2020-03-13
Ensure voting works fine with a lot of concurrent requests


hanzei:
When a lot of users are voting at the same time on the same poll, races to the data base can happen. Use `KVSetWithOptions` with `Atomic` to prevent this.

PR #276, opened 2020-03-21
Fix races for concurrent votes

hanzei:
Fixes https://github.com/matterpoll/matterpoll/issues/269


server/plugin/api_test.go
```diff
@@ -178,7 +178,7 @@ func TestHandleCreatePoll(t *testing.T) {
 		request := &model.PostActionIntegrationRequest{UserId: "userID1", TeamId: "teamID1"}
 
 		w := httptest.NewRecorder()
-		url := fmt.Sprintf("/api/v1/polls/create")
+		url := "/api/v1/polls/create"
```
hanzei: This is just cleanup

server/store/store.go
```diff
@@ -11,11 +13,12 @@ type Store interface {
...
+	Insert(*poll.Poll) error
+	Update(prev *poll.Poll, new *poll.Poll) error
+	Delete(*poll.Poll) error
 }
 
-// SystemStore allows to access system informations in the store.
+// SystemStore allows to access system information in the store.
 type SystemStore interface {
 	GetVersion() (string, error)
 	SaveVersion(version string) error
```
hanzei: I didn't touch these two as they are only run once on a node and multi node coordination would be needed.

server/store/store.go
```diff
@@ -11,11 +13,12 @@ type Store interface {
 // PollStore allows the access polls in the store.
 type PollStore interface {
 	Get(id string) (*poll.Poll, error)
-	Save(poll *poll.Poll) error
-	Delete(poll *poll.Poll) error
+	Insert(*poll.Poll) error
+	Update(prev *poll.Poll, new *poll.Poll) error
+	Delete(*poll.Poll) error
```
hanzei: I thought about if a `prev` variable is needed for `Delete`, but in the end every time you do delete, you want to do it anyway.

Issue #282, opened 2020-04-04
Voting is not possible if two users have already voted


hanzei:
**This only effects `master`**

1. User A votes for Option 1
2. User B votes for Option 1
3. User A votes for Option 2

Step 3. fails with: `WARN[2020-04-04T10:07:38.8020265+02:00] failed to handle PostActionIntegrationRequest  caller="mlog/sugar.go:27" error="failed to save poll: poll already exists in database" plugin_id=com.github.matterpoll.matterpoll
` in the server logs.

This bug got introduced with https://github.com/matterpoll/matterpoll/pull/276.

PR #283, opened 2020-04-04
Deep copy Voters

hanzei:
`Poll.AnswerOption.Voter` was't correctly deep copied and therefore changes on the origin also changed the copy. This PR fixed it by correctly deep copying `Voters`.

Another carve is that `Voter` is `nil` for "new" options, but might become `[]string{}` when a user first votes for it and then votes for another option. To avoid this confusion in the future, `[]string{}` is now used for new options. Please note that polls fetched from the DB might still have a `nil` value, hence we have to still think about this case in the future.

Fixes https://github.com/matterpoll/matterpoll/issues/282


hanzei:
Good suggestion @kaakaa :+1:

Issue #284, opened 2020-04-05
Specify ReleaseNotesURL


hanzei:
Specify `ReleaseNotesURL` to be used in the Marketplace. See https://github.com/mattermost/mattermost-server/pull/13677 for more details.

PR #285, opened 2020-04-05
Specify ReleaseNotesURL

hanzei:
Fixes https://github.com/matterpoll/matterpoll/issues/284


plugin.json
```diff
@@ -4,6 +4,7 @@
   "description": "Polling feature for Mattermost's custom slash command",
   "homepage_url": "https://github.com/matterpoll/matterpoll",
   "support_url": "https://github.com/matterpoll/matterpoll/issues",
+  "release_notes_url": "https://github.com/matterpoll/matterpoll/releases/tag/v1.3.0",
```
kaakaa, line 7:
```suggestion
  "release_notes_url": "https://github.com/matterpoll/matterpoll/releases/latest",
```

With `/releases/latest`, we won't have to update this link with each release.

https://help.github.com/en/github/administering-a-repository/linking-to-releases
> Alternatively, right click Latest Release and copy the URL to share it. The suffix of this URL is always /releases/latest.

hanzei: Using `https://github.com/matterpoll/matterpoll/releases/latest` might cause issues:
1. `v1.4.0` get released with `release_notes_url` set to `https://github.com/matterpoll/matterpoll/releases/latest`. `release_notes_url` is shown every time a users updated to `v1.4.0`.
2. Once `v1.5.0` gets released users updating to `v1.4.0` would see `https://github.com/matterpoll/matterpoll/releases/latest` as the release notes, which incorrectly point to `https://github.com/matterpoll/matterpoll/releases/v1.5.0`

kaakaa: Got it! Thanks. We'll need to keep in mind that we'll update this field in release task.

PR #290, opened 2020-04-25
Improve input parser

McKael:
The aim of this PR is to improve the input parser, which has many shortcomings.

Main changes:

* Multiple space characters don't break the parser
* The options can be placed in any position
* Quotes can be omitted (as long as at least one argument is quoted)

For example, it is now possible to use:

    /poll --progress "Is Matterpoll great?" Yes No Maybe

There is a backward-incompatible change: `"TEXT"--param1--param2` is no more split into an option "TEXT" and two parameters "param1" and "param2" (which made no sense IMHO...).

The tests have been added/updated.

## Summary by CodeRabbit

- **New Features**
  - Improved input parsing so user entries handle quotes, escape sequences, backslashes, and varied whitespace more reliably; single unquoted entries are treated as a single question and trimming of surrounding whitespace is applied.

- **Tests**
  - Expanded test suite with many edge cases (escaped spaces, complex quoting, backslashes, whitespace trimming, no-options scenarios) to ensure consistent parsing.


kaakaa:
Thanks @McKael. Your points are good, and your implementation is working fine.

And I agree `Multiple space characters don't break the parser` should be fixed, however, I don't think the others should necessarily be changed. And I feel some pain to maintain sequential parsing. So wouldn't it be possible to improve by extending the existing implementation?

If you have a specific reason for this improvement, please tell us more about it.

hanzei:
One site note: The upcoming version will allow users to create polls using a modal opened by `/poll`. Hence, I'm hesitant to do more changes to the "old" poll creation style.

McKael:
@kaakaa The reason I changed the implementation is that the current one (splitting arguments on `"<space>"`) isn't reliable and can fail in other (contrived) corner cases.
For example, parsing `/poll "A \" " "B" "C"` would fail as well.
It would also badly parse something like ` /poll A "B" "C" --abc`, BTW.

Initially I just wanted to fix the multiple-space issue, but splitting on several space characters didn't seem trivial either so I thought we could make the whole parser more robust.
Once it's working I don't think it really add pain to maintain, since it's self-contained...

Besides, I really like the ability to omit unnecessary quotes and to put flags first like in any usual command line...

@hanzei `/poll` brings a pop-up window but is the command line usage going to be removed?  (I hope not!)

Thanks for your feedback :)

hanzei:
@McKael Heads up that I still have this PR on my review list. I will get back to it as soon as I can.

jl386:
@McKael Could you look to incorporate https://github.com/matterpoll/matterpoll/issues/376 into these improvements?

hanzei:
@McKael Do you want to continue to work on the PR?

patrickmelix:
Even though old, could this be merged to get rid of the annoying problem https://github.com/matterpoll/matterpoll/issues/558 ?

hanzei (changes requested):

hanzei:
Closing as stale. Still, thanks for the PR!

Issue #291, opened 2020-04-26
Hide inactive buttons


globalqss:
When a new poll is created it shows the three buttons (Add option, Delete poll, End poll) to everybody.

![buttonspoll](https://user-images.githubusercontent.com/744634/80310051-f2896e00-87d8-11ea-88c9-74a5ef71f52e.png)

When normal users try to use them, the message is shown:
* Only the creator of a poll and System Admins are allowed to add options.  (UNLESS you have created the poll with --public-add-option)
* Only the creator of a poll and System Admins are allowed to delete it.
* Only the creator of a poll and System Admins are allowed to end it.

From UX perspective the buttons that cannot be used must be hidden - or at least clearly inactive (dimmed and not clickable).

kaakaa:
commented on https://github.com/matterpoll/matterpoll/issues/289#issuecomment-621590705.

hanzei:
Closing as duplicate https://github.com/matterpoll/matterpoll/issues/291

PR #292, opened 2020-04-30
Show "Add Option" button for only users who have permission

### LINKED ISSUE #288: Default behavior and buttons don't seem to match options
#### Summary
The default addionnal buttons specifically the Àdd Option` one is always on no matter if the `--public-add-option` is used or not.

#### Steps to reproduce
1. start a poll with `/poll "q" "a" "b"`
2. the following popup shows up for **ALL** users
image

I wouldn't want the users to be able to add an option by default and i didn't use the  `--public-add-option` option. 

#### Expected behavior
Users shouldn't see that button

Add option button shows up

#### Server configuration
**Operating system:**
Linux

**Mattermost version:**
5.21.0

**Matterpoll version:**
1.3.0

**Updated from an older Matterpoll version or fresh install:**
Fresh Install

kaakaa:
## Summary
This PR makes "Add Option" button visible for only users who have permission to add options. "Add Option" button is shown for the following users.

* system administrator
* poll creator
* (only if `--public-add-option` set) all users

スクリーンショット 2020-04-29 9 54 25

Fixes #288, #291


server/poll/metadata.go
```diff
@@ -2,18 +2,20 @@ package poll
 
 // Metadata stores personalized metadata of a poll.
 type Metadata struct {
-	PollID          string   `json:"poll_id"`
-	UserID          string   `json:"user_id"`
-	AdminPermission bool     `json:"admin_permission"` // AdminPermission will be true if the user with "UserID" has admin permission for the poll with "PollID", otherwise false.
-	VotedAnswers    []string `json:"voted_answers"`    // VotedAnswers is list of answer that the user with "UserID" have voted for the poll with "PollID"
+	VotedAnswers           []string `json:"voted_answers"` // VotedAnswers is list of answer that the user with "UserID" have voted for the poll with "PollID"
+	PollID                 string   `json:"poll_id"`
+	UserID                 string   `json:"user_id"`
```
hanzei, line 7: Out of scope of this PR: Could you please elaborate why `userID` is needed here?

server/poll/metadata.go
```diff
@@ -2,18 +2,20 @@ package poll
 
 // Metadata stores personalized metadata of a poll.
 type Metadata struct {
```
hanzei, line 4: 0/5 thought, which it out of scope.: It feels like `Metadata` and `Poll` are very similar. Maybe they can be unified at some point.

server/poll/metadata.go
kaakaa: There is no practical reason, but I think that `pollID` and `userID` are primary key for `Metadata`.

server/poll/metadata.go
kaakaa: I thought the same thing, and I tried to add `poll.Settings` field to `Metadata`. But the only data types that can be handled between the server and webapp are those specified by gob [here](https://github.com/mattermost/mattermost-server/blob/e4fb5791b0ccfe22d2e2c8bdb7df5a410675f0a7/plugin/client_rpc.go#L95), so specifying the `poll.Settings` field resulted in an encoding error. As a results, we may not be able to use `Poll` type as `Metadata`.

server/poll/metadata.go
hanzei: It makes sense that `pollID` and `userID` are the primary key when fetching `Metadata`, but I'm wondering why they are also get returned? A client should how the `pollId` and `userID` itself.

PR #304, opened 2020-06-14
Set voter count to Metadata when progress option was set

### LINKED ISSUE #302: With progress settings, button color never be changed
#### Summary
スクリーンショット 2020-06-05 0 34 54

#### Steps to reproduce
1. Enable ExperimentalUI for Matterpoll
2. Create poll with progress settings.  e.g.: `/poll "test" --progress`
3. Vote an answer

#### Expected behavior
The color of button that user voted was changed

The color doesn't change

#### Server configuration
**Operating system:** macOS

**Mattermost version:** master branch

**Matterpoll version:** 84_multi branch

**Updated from an older Matterpoll version or fresh install:** fresh

#### Additional context
With progress settings, post_action.name of buttons for poll option has suffix `XXX (0)`. But matterpoll doesn't handle it.
https://github.com/matterpoll/matterpoll/blob/f0560e5595f28a243cad77bd976c10f189d572a7/webapp/src/components/post_type/action_view/action_view.jsx#L53

kaakaa:
fixes #302

### DIFF
--- a/server/poll/poll.go
+++ b/server/poll/poll.go
@@ -133,9 +133,13 @@ func (p *Poll) getVotedAnswers(userID string) ([]string, error) {
 	}
 	votedAnswer := []string{}
 	for _, o := range p.AnswerOptions {
+		answer := o.Answer
+		if p.Settings.Progress {
+			answer = fmt.Sprintf("%s (%d)", answer, len(o.Voter))
+		}
 		for _, v := range o.Voter {
 			if userID == v {
-				votedAnswer = append(votedAnswer, o.Answer)
+				votedAnswer = append(votedAnswer, answer)
 			}
 		}
 	}
--- a/server/poll/poll.go
+++ b/server/poll/poll.go
@@ -131,6 +131,7 @@ func (p *Poll) getAnswerOptionName(o *AnswerOption) string {
 	if p.Settings.Progress {
 		return fmt.Sprintf("%s (%d)", o.Answer, len(o.Voter))
 	}
+	
 	return o.Answer
 }
 
--- a/server/poll/poll.go
+++ b/server/poll/poll.go
@@ -126,6 +126,14 @@ func (p *Poll) UpdateVote(userID string, index int) error {
 	return nil
 }
 
+// getAnswerOptionName returns answer option name (with voter count if progress setting is available)
+func (p *Poll) getAnswerOptionName(o *AnswerOption) string {
+	if p.Settings.Progress {
+		return fmt.Sprintf("%s (%d)", o.Answer, len(o.Voter))
+	}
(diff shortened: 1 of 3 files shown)


server/poll/poll.go
```diff
@@ -135,7 +143,7 @@ func (p *Poll) getVotedAnswers(userID string) ([]string, error) {
 	for _, o := range p.AnswerOptions {
 		for _, v := range o.Voter {
 			if userID == v {
-				votedAnswer = append(votedAnswer, o.Answer)
+				votedAnswer = append(votedAnswer, p.getAnswerOptionName(o))
```
hanzei, line 146: Why did you fix this bug on the server side? The `(X)` suffix feels like a formnating thing that the API shouldn't return.

How hard is it to fix the bug on the webapp only?

hanzei (changes requested):

server/poll/poll.go
kaakaa: Before this changes, `(X)` suffix as the part of answer text was added on the server side [here](https://github.com/matterpoll/matterpoll/pull/304/files#diff-67c60e0c08d1bd5cd27b356d9f8ef017L66), and the logic to change selected button color is to check if answer text between PostAction and Metadata is same. So it would be better to add `(X)` suffix in Metadata on the server side, which is close to where suffix in PostAction is added.

And I think it's little hard to fix this on webapp, because webapp doesn't have information on voters other than myself, and can't count the number of voters. With regexp, it might be work fine, but I feel little weird.

hanzei: I you think it's harder to fix on the webapp I'm fine with the current fix

server/poll/poll.go
```diff
@@ -126,6 +126,14 @@ func (p *Poll) UpdateVote(userID string, index int) error {
 	return nil
 }
 
+// getAnswerOptionName returns answer option name (with voter count if progress setting is available)
+func (p *Poll) getAnswerOptionName(o *AnswerOption) string {
+	if p.Settings.Progress {
+		return fmt.Sprintf("%s (%d)", o.Answer, len(o.Voter))
+	}
+	return o.Answer
```
hanzei, line 134: Nit:
```suggestion
	}
	
	return o.Answer
```

hanzei (approved):

server/poll/poll.go
kaakaa: Thanks, but this blank line seems to cause style error.
https://app.circleci.com/pipelines/github/matterpoll/matterpoll/479/workflows/e7167a1b-e996-4a3c-bed6-f82305912cd4/jobs/869/parallel-runs/0/steps/0-106

I'll revert this.

PR #315, opened 2020-08-13
add Chinese translation

Sharuru:


CONTRIBUTING.md
```diff
@@ -27,6 +27,7 @@ The localization process is defined below:
 - Japanese: [@kaakaa](https://github.com/kaakaa)
 - Polish: [@burasuk](https://github.com/burasuk)
 - Spanish: [@ernestoittig](https://github.com/ernestoittig)
+- Chinese: [@sharuru](https://github.com/sharuru)
```
kaakaa, line 30: There are two kinds of Chinese (`zh-CN` and `zh-TW`) in Mattermost, so I think that it has better to clarify which Chinese have been translated. According to [Mattermost docs](https://docs.mattermost.com/developer/localization.html#translation-maintenance), `zh-CN` is called `Simplified Chinese`, so what do you think about rewriting this and README to `Simplified Chinese`?

kaakaa (commented):
Thanks @Sharuru! Your translations works fine. But, I have a question about documentation.

CONTRIBUTING.md
Sharuru: Oh, it's my mistake.
`Simplified Chinese` would be better.

kaakaa (approved):
LGTM 👍 

@hanzei Can I merge this?

hanzei (approved):

PR #321, opened 2020-09-01
Register Command in OnActivate

hanzei:
`OnConfigurationChange` might not be called after `OnActivate`. This PR ensures that the command is correctly registered by calling `RegisterCommand` in `OnActivate`.

See https://mattermost.atlassian.net/browse/MM-27174 for more details.

### DIFF
--- a/server/plugin/plugin.go
+++ b/server/plugin/plugin.go
@@ -89,6 +89,10 @@ func (p *MatterpollPlugin) OnActivate() error {
 		return errors.Wrap(err, "failed to patch bot description")
 	}
 
+	if err := p.API.RegisterCommand(p.getCommand(p.getConfiguration().Trigger)); err != nil {
+		return errors.Wrap(err, "failed to register  command")
+	}
+
 	p.router = p.InitAPI()
 
 	p.setActivated(true)
--- a/server/plugin/plugin_test.go
+++ b/server/plugin/plugin_test.go
@@ -47,6 +47,13 @@ func TestPluginOnActivate(t *testing.T) {
 		DisplayName: botDisplayName,
 	}
 
+	command := &model.Command{
+		Trigger:          "poll",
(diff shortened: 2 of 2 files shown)


server/plugin/plugin.go
```diff
@@ -89,6 +89,10 @@ func (p *MatterpollPlugin) OnActivate() error {
 		return errors.Wrap(err, "failed to patch bot description")
 	}
 
+	if err := p.API.RegisterCommand(p.getCommand(p.getConfiguration().Trigger)); err != nil {
```
lieut-data, line 92: Should we instead be calling `OnConfigurationChange()` here? Until that's called, `p.getConfiguration()` will just be returning the nil value for the `configuration` struct, which seems like it won't yet have the configured `trigger`.

kaakaa: @lieut-data's suggestion looks good to me. But there is a concern that `OnConfigurationChange()` relies on `p.isActivated()`, and `p.isActivated()` does not return `true` until `p.setActivated(true)` in few lines later is done. So I think that `OnConfigurationChange()` should be called after `p.setActivated(true)`. (but it looks little strange)

hanzei: The MM server ensures that `OnConfigurationChange` is called before `OnActivate` https://github.com/mattermost/mattermost-server/blob/57244d06393502eba388bf8c140b8158cc3da80a/plugin/client_rpc.go#L214. I'm inclined to rely on that mechanism.

kaakaa: Thanks @hanzei. I missed that mechanism has been documented.  https://developers.mattermost.com/extend/plugins/server/reference/#Hooks.OnConfigurationChange. So I'm fine with your changes.

kaakaa (approved):

PR #327, opened 2020-10-05
Post responses in RHS

hanzei:
I shield away from adding a test for the `post.RootId != ""` case as this would require refactoring in way tests are setup.


kaakaa (approved):
I didn't know this behaviour... thanks 👍

PR #353, opened 2020-10-21
Disable dependabot

hanzei:
The notifications for new releases have become to noisy. Let's just update security issues ASPA.


.github/dependabot.yml
```diff
@@ -1,12 +1,15 @@
...
   reviewers:
-  - hanzei
+  - "hanzei"
   assignees:
-  - hanzei
+  - "hanzei"
   labels:
-  - 3. To Review
+  - "3. To Review"
+  allowed_updates:
```
kaakaa, line 13: is `allowed_updates` valid setting? [GitHub docs](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates) doesn't have any section about this setting.
`allowed_updates` is in [Dependabot Preview's docs](https://dependabot.com/docs/config-file/), but I think that matterpoll uses **GitHub**'s dependabot.

(Since I don't know much about dependabot, trying   this settings may make sense)

hanzei: Good catch, that seems to be the case. https://github.com/dependabot/dependabot-core/issues/1828 also backs this up.

Well, seems like we have to live with the noisy messages. Should we disable it then and update by hand before every release?

kaakaa: Disabling seems to be fine until `allowed_updates` or its equivalent is introduced.

kaakaa (approved):

PR #381, opened 2021-05-05

kaakaa:
Downstream changes from https://github.com/mattermost/mattermost-plugin-starter-template 

- Add `plan.yml` for sync and `make sync` with starter-template
- Disable `fieldalignment` of govet
- Change setup script of webapp test from `js` to `tsx`


build/manifest/main.go
```diff
@@ -114,7 +114,7 @@ func applyManifest(manifest *model.Manifest) error {
 
 		// write generated code to file by using Go file template.
 		if err := ioutil.WriteFile(
-			"server/plugin/manifest.go",
+			"server/manifest.go",
```
hanzei, line 117: Is that change intended? `server/manifest.go` doesn't seem to be checked in.

webapp/tests/setup.tsx
```diff
@@ -1,11 +1,14 @@
-import '@babel/polyfill';
-import Adapter from 'enzyme-adapter-react-16';
-import {configure} from 'enzyme';
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
```
hanzei, line 2: Let's drop these lines

hanzei (changes requested):
Thanks for keeping the plugin up to date :+1:

build/manifest/main.go
kaakaa: Good catching 👍  Intended, but this doesn't seem to make sense. I revert it.

hanzei (approved):

PR #382, opened 2021-05-09
Add reset-vote button when a poll doesn't have multi votes setting

### LINKED ISSUE #379: Allow a user to remove a cast vote.
#### Summary

It would be helpful to allow users to change their mind.  In particular, a user may have cast a vote by mistake: they didn't mean to take part in the survey.  It would be helpful if the user had the possibility to remove their vote.

The only alternative that I know of would be to remove the poll and start from scratch, asking users to re-register their votes.

#### Additional context

There's no other context that I know of.

kaakaa:
fixes #379 

From #379's story, Matterpoll would not suppress showing reset button even if a poll doesn't have multi vote settings.

About effects for existing polls, when someone performs any action on a poll, the reset button in the poll will be displayed for all users. The reset button works fine.


server/poll/transform.go
```diff
@@ -53,21 +53,19 @@ func (p *Poll) ToPostActions(localizer *i18n.Localizer, pluginID, authorName str
...
-				Type: model.POST_ACTION_TYPE_BUTTON,
-				Integration: &model.PostActionIntegration{
-					URL: fmt.Sprintf("/plugins/%s/api/v1/polls/%s/votes/reset", pluginID, p.ID),
-				},
+	actions = append(actions,
+		&model.PostAction{
+			Id: "resetVote",
+			Name: localizer.MustLocalize(&i18n.LocalizeConfig{DefaultMessage: &i18n.Message{
+				ID:    "poll.button.resetVotes",
+				Other: "Reset Votes",
```
hanzei, line 61: Maybe this should be `Reset your Votes`?

Also: Should it say `Vote` if there is not  multi votes setting?

hanzei (commented):
Thanks for the PR :+1: Two questions

server/poll/transform.go
kaakaa: Good catching! Fixed 👍

hanzei (approved):

PR #404, opened 2022-01-16
Drop `MustLocalize` and use `mattermost-plugin-api/i18n` instead of `nicksnyder/go-i18n` directly

### LINKED ISSUE #260: Move i18n code into own package
This helps to isolate the code more and get allows us to drop `localizer.MustLocalize`, which caused https://github.com/matterpoll/matterpoll/issues/252

kaakaa:
fixes #260 

## Notes
- Add linters to limit the use of `github.com/nicksnyder/go-i18n`(depguard) ~~and `MustLocalize`(forbidigo)~~
- Why not use latest version of mattermost-plugin-api?
  - mattermost-plugin-api@v0.0.20 or later is work with mattermost v6 or later, but v5.37(ESR) still be supported. So mattermost-plugin-api@v0.0.19 is the best for now


server/plugin/api.go
```diff
@@ -548,19 +548,21 @@ func (p *MatterpollPlugin) handleAddOptionConfirm(vars map[string]string, reques
...
+	userLocalizer := p.bundle.GetUserLocalizer(poll.Creator)
 
 	if errMsg := poll.AddAnswerOption(answerOption); errMsg != nil {
 		response := &model.SubmitDialogResponse{
 			Errors: map[string]string{
-				addOptionKey: p.LocalizeErrorMessage(userLocalizer, errMsg),
+				addOptionKey: p.bundle.LocalizeWithConfig(userLocalizer, &i18n.LocalizeConfig{
+					DefaultMessage: errMsg.Message,
+					TemplateData:   errMsg.Data,
+				}),
```
hanzei, line 559: Maybe this should be wrapped into helper method?

hanzei (approved):
Nice cleanup :+1:

server/plugin/api.go
kaakaa: Good catching, thanks @hanzei 👍

PR #405, opened 2022-01-16
Drop bou.ke/moneky

kaakaa:
Since [bou.ke/monkey's license](https://github.com/bouk/monkey/blob/master/LICENSE.md) says that `I do not give anyone permissions to use this tool for any purpose.`, drop `bou.ke/monkey` and use [`undefinedlabs/go-mpatch`](https://github.com/undefinedlabs/go-mpatch) (released under MIT license) instead.

`undefinedlabs/go-mpatch` seems to be used by [Argo](https://github.com/argoproj/argo-cd/blob/master/go.mod#L67) which is the CNCF incubating project.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/matterpoll/matterpoll
 go 1.12
 
 require (
-	bou.ke/monkey v1.0.2
 	github.com/blang/semver/v4 v4.0.0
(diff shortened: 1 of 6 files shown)


hanzei (approved):
Thanks for the cleanup :+1:

PR #427, opened 2022-05-17
Poll options to allow anyone to delete or end a poll

timkral:
Hi,

We have created a Mattermost integration in which users invoke actions and responses from a bot account via a slash command. One such action we would like to include is the ability for the bot account to assemble and post polls. In other words, the custom slash command becomes shorthand for creating what otherwise would be very verbose poll commands.

The issue is that we use the bot account personal access token to interact with the Mattermost server. In the case of polls, we are using the bot token to execute the poll command. This works fine, however, because the bot account is the one ultimately creating the poll, no one outside of admins are able to delete or end it due to the permission checks guarding those actions.

We'd like to find a way around this permission issue to allow the bot account to create the poll, but a non-admin user to delete or end it. It turns out, in this narrow use case, we are actually fine with anyone deleting or ending the poll and so this PR introduces two new poll command options: `--public-delete-poll` and `--public-end-poll`. Both follow the exact same pattern laid out by the `--public-add-option` flag.

Thanks for the consideration. Please let me know if there's anything that I can/should update.
-Tim Kral

### DIFF
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -577,12 +577,14 @@ func (p *MatterpollPlugin) handleEndPoll(vars map[string]string, request *model.
 		return &i18n.LocalizeConfig{DefaultMessage: commandErrorGeneric}, nil, errors.Wrap(err, "failed to get poll")
 	}
 
-	canManagePoll, appErr := p.CanManagePoll(poll, request.UserId)
-	if appErr != nil {
-		return &i18n.LocalizeConfig{DefaultMessage: commandErrorGeneric}, nil, errors.Wrap(appErr, "failed to check permission")
-	}
-	if !canManagePoll {
-		return &i18n.LocalizeConfig{DefaultMessage: responseEndPollInvalidPermission}, nil, nil
+	if !poll.Settings.PublicEndPoll {
+		canManagePoll, appErr := p.CanManagePoll(poll, request.UserId)
+		if appErr != nil {
+			return &i18n.LocalizeConfig{DefaultMessage: commandErrorGeneric}, nil, errors.Wrap(appErr, "failed to check permission")
+		}
+		if !canManagePoll {
+			return &i18n.LocalizeConfig{DefaultMessage: responseEndPollInvalidPermission}, nil, nil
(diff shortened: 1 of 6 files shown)


kaakaa:
Thank you for your suggestions @timkral 
Since CI has been in error for now, I'll review this after fixing CI error.

kaakaa:
Thank you for your investigation @timkral. Yes, this failure doesn't seem to be caused by any code changes (of course including your changes), caused by dependencies or CI environment.

This issue is discussed in [Mattermost community server](https://community-daily.mattermost.com/core/channels/extensibility-meeting/h3gdpbm5gpdai8xxmk1jc4k3ro), and the solution seems to be settling on changing the CI script.

Please wait a little longer as I'll reach out you after fixing CI issue, and ask you to merge master branch.

kaakaa:
Before reviewing in details, I would like to ask you to implement two more parts for adding new poll settings. If it's hard for you to implement these, please let me know it.

1. New settings should be implemented on interactive dialog for creating poll. (Interactive dialog is showed by executing `/poll` command without any args.)
    - Build [interactive dialog](https://developers.mattermost.com/integrate/admin-guide/admin-interactive-dialogs/)
      - https://github.com/matterpoll/matterpoll/blob/d2b1878be7ce876cb77500688f4ef36706106c5f/server/plugin/command.go#L232
    - Handle http request from interactive dialog
      - https://github.com/matterpoll/matterpoll/blob/d2b1878be7ce876cb77500688f4ef36706106c5f/server/plugin/api.go#L282

2. New settings also should be handled by webapp side. Webapp code is used if `Experimental UI` turns on.
    - Control whether the button is displayed or not
      - https://github.com/matterpoll/matterpoll/blob/b90bd0f0c7b83f311549b9ff8180d3cad74edd71/webapp/src/components/post_type/action_view/action_view.jsx#L60

hanzei:
@timkral What do you think about the feedback from @kaakaa? Do you want to continue to work on the ticket?

hanzei:
Closing an inactive

Issue #96, opened 2018-10-30
Show how voted during poll via Poll Setting


JtheBAB:
Hi

My users ask me if it would be possible to see how voted for which answer when the poll is active. At the moment there is "just" the numbers per answer visible.

Best regards

JtheBAB

hanzei:
While I see the positive sides of such Poll Setting, the UX can be tricky. How would you such information?
Currently if you use `progress` the number of people is shown in the button. Would you also show the usernames there?

lkraav:
> Currently if you use progress the number of people is shown in the button. Would you also show the usernames there?

@hanzei probably yes.

For scheduling "poll" use cases, ere's a screenshot example of a display format that really works well (from doodle.com). Obviously this exact fidelity level is probably out of reach here, at least to start, but perhaps there's a "barebones" way of achieving a similarly understandable display of poll results.

![image](https://user-images.githubusercontent.com/147228/47748443-82f1fb80-dc93-11e8-8ae0-15f9f242817e.png)

JtheBAB:
I think in the button it will make the button to big when many people are voting. Maybe like the result but „just“ dynamic.

piotr-wegert:
It would be very nice to have this option. One of our use cases Slack before we switched to MM was to start a poll about who wants to go on which conference (with voting on multiple options allowed). 

The main benefit was that folks could see who is going where and make their decision based on that.

deedubdevs:
@hanzei Usernames don't need to be on the button. [Simple poll](https://simplepoll.rocks/) for slack did it in a functional way. 
[Video example](https://simplepoll.rocks/static/main/images/new-landing-page/basic-poll-voting.ca4162c5a9a2.webm)

hanzei:
> @hanzei Usernames don't need to be on the button. [Simple poll](https://simplepoll.rocks/) for slack did it in a functional way.
> [Video example](https://simplepoll.rocks/static/main/images/new-landing-page/basic-poll-voting.ca4162c5a9a2.webm)

I know simple poll. But they use a different approach with answer options not in the  buttons, which I don't like. I'm still looking for a design that keeps the answer options in the buttons. The responses can be outside of the buttons IMO.

hanzei:
Idea: Add "Show current results" button that shows the current list of votes in a modal.

mjnaderi:
[...]

PR #431, opened 2022-05-31
Add RHS informations card to show users votes

jbattistispiria:
image

- ~~Add a new option when creating a poll «Show Voters»~~.
- ~~Add error when anonymous option is use with show voters option~~.
- Add unit tests

Change made after comments:
- Add extra information when using the `--progress` setting

Show extra information on who voted for what during poll in the RHS feed.
The extra information is send in markdown by using a "card" props from the Mattermost API.
Mattermost post add the info icon to the post and create the RHS.

Fix https://github.com/matterpoll/matterpoll/issues/96#


kaakaa (changes requested):
Great 👍  This feature works fine for me, and your changes almost looks great! I left a few review comments and would appreciate your response.

And Let me make one suggestion. How about including the functionality of `--show-voters` in the existing `--progress` setting?  Separating those two poll settings would make `--progress` just display the number of votes on  buttons, which does not seem useful. And I don't want to increase the number of poll setting as much as possible. How does this sounds for your use case?

server/poll/transform.go
```diff
@@ -208,3 +224,41 @@ func (p *Poll) ToEndPollPost(bundle *utils.Bundle, authorName string, convert ID
...
 	return post, nil
 }
+
+// ToCard return the poll for rhs card
+func (p *Poll) ToCard(bundle *utils.Bundle, convert IDToNameConverter) string {
+	localizer := bundle.GetServerLocalizer()
+	s := fmt.Sprintf("# %s\n", p.Question)
+
+	creatorName, _ := convert(p.Creator)
+	s += fmt.Sprintf(localizer.MustLocalize(&i18n.LocalizeConfig{DefaultMessage: rhsCardPollCreatedBy})+" %s\n", creatorName)
```
kaakaa, line 234: Please use `bundle.LocalizeWithConfig` instead of `MustLocalize`, because `MustLocalize` causes panic if translation is missing.

refs: https://github.com/matterpoll/matterpoll/pull/404/files#diff-e4bd7a74772fffef524851fbf869db31032434192f155ad853eb40246915f760R61

server/plugin/api.go
```diff
@@ -331,7 +331,11 @@ func (p *MatterpollPlugin) handleCreatePoll(_ map[string]string, request *model.
 			"poll_id": poll.ID,
 		},
 	}
+
 	model.ParseSlackAttachment(post, actions)
+	if poll.Settings.ShowVoters {
+		post.AddProp("card", poll.ToCard(p.bundle, p.ConvertCreatorIDToDisplayName))
```
kaakaa, line 337: I would like to use `p.ConvertUserIDToDisplayName` instead of `p.ConvertCreatorIDToDisplayName` here, because it make a card more informative.
スクリーンショット 2022-06-16 23 01 15

kaakaa (commented):
I forgot to send comments.

jbattistispiria:
About including the functionality of --show-voters in the existing --progress setting, i thought the same thing as you at first and i did the code that way, but i wasn't sure if people using --progress would want to have the extra information sidebar.
So i keep the two settings separated.
I'm open to make the change, let me know if that's go.

server/plugin/api.go
jbattistispiria: I'm on it.

kaakaa:
Thanks for sharing your thought, and I understand it. But I still be inclined to include this functionality in `--progress`. My thought is the following.

* The card is not visible unless a user click on it, so it is not obtrusive
* A poll with `--progress` makes the voting user name public as a result, so I don't think it will be a problem if the voting user name is made public during a poll is open
* If a poll is created with `--progress` and `--anonymous`, with existing implementation, the voting user name will be hidden and anonymity is protected
  * In this case, a card doesn't have any useful information, but it would not be bad

If I seem to have missed something, let me know it.

jbattistispiria:
Great. I'll completely agree with you and i will move show-voters with progress.

server/poll/transform.go
jbattistispiria: Done

server/plugin/api.go
jbattistispiria: Done

jbattistispiria:
I made the change to include the functionality of `--show-voters` in the existing `--progress setting`.

server/plugin/command.go
```diff
@@ -194,6 +194,9 @@ func (p *MatterpollPlugin) executeCommand(args *model.CommandArgs) (string, *mod
 		},
 	}
 	model.ParseSlackAttachment(post, actions)
+	if newPoll.Settings.Progress {
+		post.AddProp("card", newPoll.ToCard(p.bundle, p.ConvertCreatorIDToDisplayName))
```
kaakaa, line 198:
```suggestion
		post.AddProp("card", newPoll.ToCard(p.bundle, p.ConvertUserIDToDisplayName))
```

server/poll/transform.go
```diff
@@ -208,3 +224,41 @@ func (p *Poll) ToEndPollPost(bundle *utils.Bundle, authorName string, convert ID
...
+				if i+1 == len(o.Voter) && len(o.Voter) > 1 {
+					voter += " " + bundle.LocalizeWithConfig(localizer, &i18n.LocalizeConfig{DefaultMessage: rhsCardPollVoterSeperator}) + " "
+				} else if i != 0 {
+					voter += comma
+				}
+				voter += displayName
+			}
+		}
+
+		s += "### " + localizer.MustLocalize(&i18n.LocalizeConfig{
```
kaakaa, line 254:
```suggestion
		s += "### " + bundle.LocalizeWithConfig(localizer, &i18n.LocalizeConfig{
```

kaakaa (commented):
Thanks 👍  There are 2 additional points that I would like to fix, so please approve them if they looks good.

server/poll/transform.go
jbattistispiria: Done, i had to make some changes to unit tests.

kaakaa (approved):
Perfect 👍

kaakaa:
Since Matterpoll is on a way to [release v1.5.0](https://github.com/mattermost/mattermost-marketplace/issues/298), I'd like to wait to merge this PR until the release process moves forward. If the release process seems to take a long time, I'll cut a new branch for `release-1.5` and merge this. I'm careful not to generate additional work for this PR.

kaakaa:
Since there is no activity for a month in https://github.com/mattermost/mattermost-marketplace/issues/298, move forward this PR. Thanks again for your contribution @jbattistispiria!

hanzei:
@jbattistispiria This is great, thanks for the contribution. :tada:

Quick q: Does this feature work on mobile?

kaakaa:
No. AFAIK, `card` props is not supported on Mobile, so RHS information cannot be seen on mobile.

physiker05:
Hi all! Although a bit outdated... I'd like to know how I get to see this card that shows me the users participating in an open poll? I started a poll using the option `--progress`, but I don't get to see the stats. Your help is very much appreciated!

kaakaa:
@physiker05 I'm sorry that I didn't catch your comment.
First let me check, what version of matterpoll are you using? If you are using Matterpoll v1.6.0 or later, you should be able to see the icon to open the card by adding the `--progress` option.

open-card

physiker05:
@kaakaa Thanks for getting back to me! Well, you probably identified the main issue... I'm running on v1.4.0 😁.

PR #432, opened 2022-06-16
Update active.de.json

vorlif:
The German translations have been updated. 
I also adjusted some old translations if there were spelling mistakes or if I thought the new translations fit better in parts.

@hanzei Want to take a look?

### DIFF
--- a/assets/i18n/active.de.json
+++ b/assets/i18n/active.de.json
@@ -37,7 +37,7 @@
   },
   "command.help.text.pollSetting.anonymous": {
     "hash": "sha1-413dccc620dcf0d910c2cba09522443793fbcb33",
-    "other": "Nach Abschluss der Umfrage nicht anzeigen, wer für was gestimmt hat"
+    "other": "Nicht anzeigen, wer wofür gestimmt hat"
   },
   "command.help.text.pollSetting.introduction": {
     "hash": "sha1-5cf3ba00c35f9b75d8b2118081c7c45b2a36ecb3",
@@ -159,7 +159,7 @@
   },
   "poll.newPoll.votesettings.invalidSetting": {
     "hash": "sha1-bf96d6b01204be0f86727d2253d235272e177ded",
-    "other": "Die Anzahl der Stimmen muss eine positive Zahl und kleiner oder gleich der Anzahl der Antworten sein. Du hast \"{{.MaxVotes}}\" angegeben, aber die Anzahl der Optionen ist \"{{.Options}}\".",
+    "other": "Die Anzahl der Stimmen muss eine positive Zahl und kleiner oder gleich der Anzahl der Antworten sein. Du hast \"{{.MaxVotes}}\" angegeben, aber die Anzahl der Optionen ist \"{{.Options}}\"."
   },
   "poll.newPoll.votesettings.unexpectedError": {
     "hash": "sha1-e9d86776544f9cf7effd6db84f9e988b6fbccd29",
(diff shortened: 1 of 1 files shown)


assets/i18n/active.de.json
```diff
@@ -157,44 +158,44 @@
     "other": "Unbekannte Umfrageeinstellung: {{.Setting}}"
   },
   "poll.newPoll.votesettings.invalidSetting": {
-    "hash": "sha1-a0649f6d7040b480c543e843e12cd14de97ef6d6",
-    "other": "Die Anzahl der Stimmen musss eine positive Zahl kleiner gleich der Anzahl der Antwortoptionen sein. Gewählt wurde \"{{.MaxVotes}}\", aber die Anzahl der Antwortoptionen ist \"{{.Options}}\"."
+    "hash": "sha1-bf96d6b01204be0f86727d2253d235272e177ded",
+    "other": "Die Anzahl der Stimmen muss eine positive Zahl und kleiner oder gleich der Anzahl der Antworten sein. Du hast \"{{.MaxVotes}}\" angegeben, aber die Anzahl der Optionen ist \"{{.Options}}\".",
```
kaakaa, line 162: Please remove a comma in the end of this line.

assets/i18n/active.de.json
```diff
@@ -25,39 +25,39 @@
...
   },
   "command.help.text.options": {
     "hash": "sha1-0b06ca6a7a2a1e2481ed761fe93eefec7d67079e",
-    "other": "Du kannst Antwortmöglichkeit festlegen, indem du `/{{.Trigger}} \"Frage\" \"Antwort 1\" \"Antwort 2\" \"Antwort 3\"` eingibst"
+    "other": "Du kannst die Optionen durch Eingabe von `/{{.Trigger}} \"Frage\" \"Antwort 1\" \"Antwort 2\" \"Antwort 3\"` anpassen"
   },
   "command.help.text.pollSetting.anonymous": {
     "hash": "sha1-413dccc620dcf0d910c2cba09522443793fbcb33",
-    "other": "Verberge, wer für was gestimmt hat"
+    "other": "Nach Abschluss der Umfrage nicht anzeigen, wer für was gestimmt hat"
```
kaakaa, line 40: I supposed that this change adds to the meaning of "after the polls ends", but I don't want to limit `--anonymous` setting to after the polls ends. Because the ongoing PR (#431) may cause the A setting to also affect the anonymity of the ongoing poll.

kaakaa (changes requested):
Thanks a lot @vorlif ! I left a few comments, and would appreciate it if you could response.
It seems like @hanzei are busy, so I'll merge this if there is no comments from @hanzei until cutting new release.

assets/i18n/active.de.json
vorlif: Yes, that's right, it means that. Of course, it also makes sense that anonymity is permanently preserved. But isn't that what the English text means? 

https://github.com/matterpoll/matterpoll/blob/master/server/plugin/command.go#L54

I believe this was translated the same way in most other translations. :thinking:

vorlif: Fixed

assets/i18n/active.de.json
vorlif: Fixed

assets/i18n/active.de.json
kaakaa: Ah...yes, you're right. I was thinking too much about #431 and the English message had slipped my mind. 
Other languages including English also should be fixed.  But I plan to release #431 as next v1.6.0 release, so this message correction could have been made in the next v1.6.0 release. I'm sorry for confusing you, and either message is fine for this release.

vorlif: No problem, thanks for your work! Then we leave the text so and with the next release, it fits then?

kaakaa: Sure 👍

kaakaa (approved):

kaakaa:
Thanks @vorlif !

PR #453, opened 2022-09-22
Use version of React DOM provided by web app

kaakaa:
I tested this with ExperimentalUI setting on Mattermost v6.3.10, and it works fine.

スクリーンショット 2022-09-22 22 54 58

refs:
* https://github.com/mattermost/mattermost-plugin-starter-template/pull/169
* https://mattermost.atlassian.net/browse/MM-47046

TODO:
- ~~[ ] It might be needed to add other packages [refs](https://github.com/mattermost/mattermost-plugin-starter-template/pull/169#issuecomment-1250414723)~~
  - This should be fixed by another PR if needed

### DIFF
--- a/webapp/webpack.config.js
+++ b/webapp/webpack.config.js
@@ -81,6 +81,7 @@ module.exports = {
     },
     externals: {
         react: 'React',
+        'react-dom': 'ReactDOM',
         redux: 'Redux',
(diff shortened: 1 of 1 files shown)


hanzei (approved):
Thanks for taking care of this @kaakaa 👍

PR #455, opened 2022-10-06
Update CI orb to latest version

hanzei:


.nvmrc
```diff
@@ -0,0 +1 @@
+13
```
kaakaa, line 1:
```suggestion
14
```

I think this should be 14, because Node 13 has already reached EoL.
https://github.com/nodejs/Release#end-of-life-releases

And EoL of Node 14 will come in next March, so we might start thinking about upgrading to Node 16 or 18. EoL of 16 will be coming in next September, so it would be great to upgrade to 18

kaakaa (approved):
Thank you for your work to update orb @hanzei .
I left a suggestion, but it doesn't block this, because it would not be a problem for building a plugin in practice.

.nvmrc
hanzei: I don't mind if someone wants to work on updating the plugin to a newer node version, but I'm hesitant to invest time into that myself. As long node 13 works, I'm fine with using it. `nvm` can install what ever version we need.

PR #457, opened 2022-10-12
Fix grammar in translation (de)

mmslkr:
Fixes a small grammar error in the german translation.


hanzei (approved):
Good catch :+1:

Und danke sehr fürs Beheben

Issue #485, opened 2023-05-12
Unable to Enable 1.6.0


opetnor:
#### Summary
Cannot enable Matterpoll 1.6.0 Air gapped

#### Steps to reproduce
1. Upload 1.6.0 in Plugin Management
2. Try to enable plugin

#### Expected behavior
We can upload plugin, but get an error.

#### Observed behavior (that appears unintentional)
What did you see happen? Please include relevant error messages and/or screenshots.

Error:
This plugin failed to start. unable to start plugin: com.github.matterpoll.matterpoll: Unrecognized remote plugin message: This usually means the plugin was not compiled for this architecture, the plugin is missing dynamic-link libraries necessary to run, the plugin is not executable by this process due to file permissions, or the plugin failed to negotiate the initial go-plugin protocol handshake Additional notes about plugin: Path: plugins/com.github.matterpoll.matterpoll/server/dist/plugin-linux-amd64 Mode: -rwxr-xr-x Owner: 998 [mattermost] (current: 998 [mattermost]) Group: 998 [mattermost] (current: 998 [mattermost]) ELF architecture: EM_X86_64 (current architecture: amd64)
#### Server configuration
**Operating system:**
Debian 11

**Mattermost version:**
7.10.0
**Matterpoll version:**
1.6.0

**Updated from an older Matterpoll version or fresh install:**
Fresh
<details>
<summary>Mattermost server logs</summary>

```
{"timestamp":"2023-05-12 09:12:20.904 +02:00","level":"debug","msg":"starting plugin","caller":"plugin/hclog_adapter.go:52","plugin_id":"com.github.matterpoll.matterpoll","wrapped_extras":"pathplugins/com.github.matterpoll.matterpoll/server/dist/plugin-linux-amd64args[plugins/com.github.matterpoll.matterpoll/server/dist/plugin-linux-amd64]"}
{"timestamp":"2023-05-12 09:12:20.905 +02:00","level":"debug","msg":"plugin started","caller":"plugin/hclog_adapter.go:52","plugin_id":"com.github.matterpoll.matterpoll","wrapped_extras":"pathplugins/com.github.matterpoll.matterpoll/server/dist/plugin-linux-amd64pid121431"}
[...]

PR #487, opened 2023-05-13
Disable CGO

kaakaa:
#485 
#### Summary
Disable CGO for all builds in order to static linking.
refs: https://github.com/mattermost/mattermost-plugin-starter-template/pull/167

This fix would be released as v1.6.1.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,12 @@ ifneq ($(wildcard build/custom.mk),)
 	include build/custom.mk
 endif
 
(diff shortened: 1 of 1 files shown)


kaakaa:
@hanzei  I believe that disabling CGO fix #485, but I would be glad to hear your thoughts if possible.

hanzei (approved):
:+1:

kaakaa:
@hanzei  Thank you so much!

Issue #500, opened 2023-09-10
Migrate plugin/plugin.go to use `github.com/mattermost/mattermost/server/public`


kaakaa:
Mattermost introduced new created go module that is for developer to use mattermost-server's public API and discretely versioned from the mattermost-server. This ticket is about migrating `plugin/plugin.go` to use that package.

This issue is a successor of #350.

kaakaa:
refs https://github.com/mattermost/mattermost-plugin-github/pull/725

PR #501, opened 2023-09-10

kaakaa:
Downstream changes from https://github.com/mattermost/mattermost-plugin-starter-template

* Bump up node version to 16.13.1
* Update webapp dependencies
* Sync eslint settings and fix lint errors
* Remove build/sync tool
* (Work to introduce `mattermost-server/server/public` will be done as a separete issue #500)


webapp/src/index.tsx
```diff
@@ -1,10 +1,15 @@
...
 
 declare global {
     interface Window {
         registerPlugin(id: string, plugin: MatterPollPlugin): void
     }
+
+    // fix for a type problem in webapp as of 6dcac2
+    type DeepPartial<T> = {
+        [P in keyof T]?: DeepPartial<T[P]>;
+    }
```
hanzei, line 12: Why is this needed?

hanzei (approved):
This is awesome :rocket: Thanks!

webapp/src/index.tsx
kaakaa: Thanks for sharing your time to review this!

> Why is this needed?

The direct reason is resolving a type error.

```
> tsc

node_modules/mattermost-webapp/packages/mattermost-redux/src/client/client4.ts:2870:27 - error TS2304: Cannot find name 'DeepPartial'.

2870     patchConfig = (patch: DeepPartial<AdminConfig>) => {                               ~~~~~~~~~~~
```

The potential reason is the issue from the version (6dcac29) of `mattermost-webapp` on which this PR depends, and it seems to be fixed by [this PR](https://github.com/mattermost/mattermost-webapp/pull/10228#discussion_r857996158).

I guess that you want to ask "Why did you use that version (6dcac29) of `mattermost-webapp` with an issue? You should use the commit including the fix of the issue.". I don't have a strong reason to use 6dcac29, but I think 6dcac29 is better to use because ...

* Fixed version with [the PR](https://github.com/mattermost/mattermost-webapp/pull/10228#discussion_r857996158) introduces other type errors.
* The commit (6dcac29) is used some other plugins such as [calls](https://github.com/mattermost/mattermost-plugin-calls/blob/main/webapp/package.json#L59), [wrangler](https://github.com/gabrieljackson/mattermost-plugin-wrangler/blob/master/webapp/package.json#L64), [channel-notes](https://github.com/crspeller/mattermost-plugin-channel-notes/blob/master/webapp/package.json#L56).
* I tried to use latest `matermost-webapp` at first, but it seemed to need a lot of work to resolve errors (related to the significant increase in Node version from 14 to 18, etc.).

So I decided to use the commit (6dcac29), which seems to have the least impact. I will use latest commit of mattermost(webapp) in next time, and remove the declaration of `DeepPatrial` at that time.

Issue #372, opened 2021-02-22
Allow a higher number of votes than options at creation time


Manuzor:
Currently, there's no way (that I'm aware of) that would allow users to add options and to have everyone able to vote for all options. The number of votes per user is currently set at creation time of the poll and is capped at the number of options at that time. However, when a new option is added later by someone else, the maximum number of votes per user stays the same.

With the experimental GUI this means that right now the maximum number of votes allowed is 3 since there's no way to add more than 3 options at creation time.

I suggest removing the upper limit on votes altogether. This would allow the creator of the poll to just specify `--votes=99` and be done with it.

I suppose this the function that would have to change:
https://github.com/matterpoll/matterpoll/blob/1f381a508314e737f7e6d1de59aed968ab3d937f/server/poll/poll.go#L164-L179

OnlineCop:
There might be a few ways to handle this. Assuming the following for illustration:

- the poll was created with 4 default options: `/poll "Should users add options?" "Yes" "I guess" "Better not" "No" --public-add-option`
- two options were submitted by users: "On the fence", "HR would disapprove"

1. `--votes=n` could allow `n <= 0`:
    - **`n>0`** indicates the hard limit of _all_ options, whether default or user-submitted.
    - **`n=0`** indicates "max" votes: users may vote for any/all default or user-submitted options.
    - **`n<0`** indicates the hard limit of votes for _default_ options (its **absolute value** would be used, examples below), but users receive one additional vote for each user-submitted option.
        - Given `--votes=0`, users have 6 votes: 4 for the default `["Yes" "I guess" "Better not" "No"]` options, plus 2 for the user-submitted `["On the fence", "HR would disapprove"]` options.
        - Given `--votes=-1`, users may only select 1 from the default options, but may additionally select one or both from the user-submitted options.
        - Given `--votes=-2`, users may select 2 from the first default options, and one or both from the user-submitted options.
    - This has the disadvantage that the poll creator couldn't limit votes for user-submitted options: if there were 10 user-submitted options, a user would have `abs(p.Settings.MaxVotes)` votes for _default_ options plus 10 more votes for each of the user-submitted ones.

2. `--votes=n` could allow `n = 0`, and add `--public-add-option-vote=m` where `m >= 0`:
    - Users would only be allowed to use these votes toward _user-submitted_ options, not toward _default_ options.
    - **`n=0`** indicates "max" votes for _default_ options, but does not affect votes toward _user-submitted_ options.
[...]

PR #506, opened 2023-10-09
Implement --votes=0 setting to vote all options  even if adding an option

### LINKED ISSUE #496: Possibility to vote for all independent of the number of options
#### Summary

It would be nice to have a multiple-response poll which allows users to vote for all options even if the number of options in the poll is updated. Currently, if a multi-response poll is created via `--vote=X` (with `X` being the number of options), once a new option is added, a user who has spent all of their `X` quota, can never vote for the new option.

I'm a mere user with no experience in developing Mattermoist plugins. So, I have not tried anything personally, but these two different approaches come to my mind:

1. Extend the `--vote=X` feature to increase the maximum number of votes `X` by one as options are added (not recommended because it may not be the intended purpose), 
2. Introduce a new boolean argument (something like `--vole-all`) that separates the `--vote=X` functionality from the vote for all feature.

#### Additional context

#256 is relevant, but a different feature.

kaakaa:
fixes #496 
refs #372

`--votes=0` setting allows users to vote all options even after adding an option. 

matterpoll-496

**Notes**: When creating a poll from interactive dialog, `0` cannot be set to `Number of Votes` due to [mattermost's issue](https://github.com/mattermost/mattermost/issues/24817). Until resolving it, `--votes=0` setting can be used only from `/poll` command.


README.md
```diff
@@ -40,7 +40,7 @@ Poll Settings provide further customisation, e.g. `/poll "Is Matterpoll great?"
 - `--anonymous`: Don't show who voted for what at the end
 - `--progress`: During the poll, show how many votes each answer option got and, in post card, show who voted for which answers ([#431](https://github.com/matterpoll/matterpoll/pull/431))
 - `--public-add-option`: Allow all users to add additional options
-- `--votes=X`: Allow users to vote for X options
+- `--votes=X`: Allow users to vote for X options. Default is 1. If X is 0, users can vote for all options even after adding options.
```
hanzei, line 43: I wonder if the wording would be more clear if it's said that the number of votes is unlimited. 
```suggestion
- `--votes=X`: Allow users to vote for X options. Default is 1. If X is 0, users have an unlimited amount of votes.
```

server/plugin/command.go
```diff
@@ -64,7 +64,7 @@ var (
 	}
 	commandHelpTextPollSettingMultiVote = &i18n.Message{
 		ID:    "command.help.text.pollSetting.multi-vote",
-		Other: "Allow users to vote for X options",
+		Other: "Allow users to vote for X options. Default is 1. If X is 0, users can vote for all options even after adding options.",
```
hanzei, line 67: See above

server/plugin/api.go
```diff
@@ -392,10 +392,15 @@ func (p *MatterpollPlugin) handleVote(vars map[string]string, request *model.Pos
...
 	}
 
+	// Multi Answer Mode
 	if poll.IsMultiVote() {
-		// Multi Answer Mode
+		var remains int
 		votedAnswers := poll.GetVotedAnswers(userID)
-		remains := poll.Settings.MaxVotes - len(votedAnswers)
+		if poll.Settings.MaxVotes == 0 {
+			remains = len(poll.AnswerOptions) - len(votedAnswers)
```
hanzei, line 400: 0/5 do we even want to show the number of remaining votes?

server/poll/poll.go
```diff
@@ -175,7 +175,7 @@ func (p *Poll) validate() *utils.ErrorMessage {
 
 // IsMultiVote return true if poll is set to multi vote
 func (p *Poll) IsMultiVote() bool {
-	return p.Settings.MaxVotes > 1
+	return p.Settings.MaxVotes == 0 || p.Settings.MaxVotes > 1
```
hanzei, line 178: Updating the code is easy :100:

server/poll/transform.go
```diff
@@ -162,7 +162,12 @@ func (p *Poll) makeAdditionalText(bundle *utils.Bundle, numberOfVotes int) strin
...
 	}
-	if p.Settings.MaxVotes > 1 {
+	switch p.Settings.MaxVotes {
+	case 0:
+		settingsText = append(settingsText, "votes=all")
+	case 1:
+		break
+	default:
 		settingsText = append(settingsText, fmt.Sprintf("votes=%d", p.Settings.MaxVotes))
 	}
```
hanzei, line 172: :100:

hanzei (changes requested):
Should we bump the minimum server version to v9.3.0 so it includes https://github.com/mattermost/mattermost/pull/24916?

server/poll/poll.go
kaakaa: Exactly 😉

kaakaa:
It seems to me that v9.3 is too new and I concerned that a number of users could be unable to launch Matterpoll v1.7. https://github.com/mattermost/mattermost/pull/24916 occurs only in dialog, not in command, so ESR v8.1 (or v7.8) looks better.

server/poll/poll.go
```diff
@@ -355,8 +355,13 @@ func (s Settings) String() string {
...
 		settingsText = append(settingsText, "public-add-option")
 	}
-	if s.MaxVotes > 1 {
-		settingsText = append(settingsText, fmt.Sprintf("votes=%d", s.MaxVotes))
+	if s.MaxVotes != 1 {
+		switch s.MaxVotes {
+		case 0:
+			settingsText = append(settingsText, "votes=all")
+		default:
+			settingsText = append(settingsText, fmt.Sprintf("votes=%d", s.MaxVotes))
```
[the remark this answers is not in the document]
kaakaa: For all settings? If so, I don't feel that just translating it would be very effective, so if you are going to do it, you may need to revamp this part of the UI as a whole. 0/5
image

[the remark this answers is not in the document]
hanzei: Good point. Let's keep that idea out of this PR.

hanzei (approved):
LGTM :+1:

Issue #511, opened 2023-10-30
Received invalid response from the server (team-edition v9.2)


braun-steven:
#### Summary
When I try to install the latest version of this plugin (1.6.1) on our Mattermost server (team-edition v9.2) by uploading the tar.gz file I get a "Received invalid response from the server." message.

#### Steps to reproduce
1. Use MM version 9.2
2. Upload matterpoll plugin version 1.6.1
3. See error

#### Expected behavior
File upload installs the plugin.
 
#### Observed behavior (that appears unintentional)
File upload fails with the following message:

<img width="956" alt="Screenshot 2023-10-30 at 07 34 38" src="https://github.com/matterpoll/matterpoll/assets/17625341/42190bb5-9ba9-40a0-9390-03f5bcf2cb4b">

#### Server configuration
**Operating system:** MacOS 14.0

**Mattermost version:** 9.2

**Matterpoll version:** 1.6.1

**Updated from an older Matterpoll version or fresh install:** fresh 

<details>
<summary>Mattermost server logs</summary>
Empty -- the file upload did not produce any logs on level: DEBUG.
</details>

kaakaa:
Thank you for reporting @braun-steven .
I would like to confirm that the `Maximum file size` in your Mattermost is set to more than the size of Matterpoll bundle? When its settings is lower than the bundle size, I can reproduce this issue. And if you use a proxy such as nginx, its maximum size settings may also be affected.
[Environment configuration settings - Maximum file size](https://docs.mattermost.com/configure/environment-configuration-settings.html#maximum-file-size)
![スクリーンショット 2023-10-31 22 17 11](https://github.com/matterpoll/matterpoll/assets/1453749/01455786-6fc6-4546-a688-69535a411ee9)


If its setting looks fine, can you install other plugin (e.g.: [matermost-plugin-demo](https://github.com/mattermost/mattermost-plugin-demo/releases)) other than Matterpoll?

braun-steven:
That solved it -- thanks! I thought it may be worth mentioning in the readme (see PR #512) since the mattermost error message is not really helpful.

PR #512, opened 2023-10-31
Add plugin file size upload issue solution in README.md

braun-steven:
A possible issue as reported in #511 is that the server settings for maximum file size is smaller than the plugin file, the upload will fail with the message "Received invalid response from the server.".

This PR adds this issue and its solution to the Troubleshooting section of the README.md.


hanzei (approved):
LGTM :+1:  

Thanks @braun-steven

PR #517, opened 2023-11-11
Show the number of voters besides total votes when `--votes=X` is enabled

### LINKED ISSUE #388: See number of unique voters while the poll is in progress
#### Summary

We need to know how many people have casted vote before closing the voting session

There is no alternative as of now, we have to end the poll to see how many people have voted

#### Additional context

Add an option to see the number of unique voters, so we may be able to take an informed choice of ending the poll when more than X percentage of voters have casted their votes

### LINKED ISSUE #515: Total votes should show both # of votes and # of voters
#### Summary

Currently, a multi-vote poll shows the number of votes, not the number of respondents. As a result, if you forget `--progress` or otherwise don't want that, you can't actually tell how many people have responded.

Instead of saying "Total votes: 29", it should say "Total responders: 20, Total votes: 29" or "Total responders: 20 (29 votes total) something to that effect.

Deleting and recreating poll w/ `--progress`, but this would lose existing data

kaakaa:
closed #388 
closed #515

Show the number of voters when `--votes=X` is enabled. Without `--votes=X` setting, this additional information is not shown because the number of voters is same as total votes.

image


server/poll/transform.go
```diff
@@ -64,10 +72,12 @@ var (
 func (p *Poll) ToPostActions(bundle *utils.Bundle, pluginID, authorName string) []*model.SlackAttachment {
 	localizer := bundle.GetServerLocalizer()
 	numberOfVotes := 0
+	voters := mapset.NewSet[string]()
```
hanzei, line 75: I'm not to keen on using a new library. Can we use a `map[string]struct{}` instead and take its `len` later?

hanzei (changes requested):
Nice work on the tests :+1: 

One blocking question about the external library, though.

server/poll/transform.go
kaakaa: It's good way, thank you for catching.

hanzei (approved):
Nice work, thanks @kaakaa :+1:

Issue #518, opened 2023-11-11
Migration CI to GitHub Actions (mattermost/actions-workflow)


kaakaa:
Need to migrate CI service to GitHub Actions from the following reason.
* [mattermost/circleci-orb](https://github.com/mattermost/circleci-orbs) supports up to go 1.19
  * and that repo doesn't seem to be active no longer
* Mattermost official plugins use [mattermost/actions-workflows](https://github.com/mattermost/actions-workflows)

- [ ] add GitHub Actions's workflow file
- [ ] change project settings (e.g.: branch protection policy, webhook, ...)

PR #519, opened 2023-11-11
Add GitHub Actions workflow file

kaakaa:
refs #518 

Add workflow file and fix some lint error.

Notes: I removed `master` branch from ci trigger because `pull_request` trigger  is always run before merging code to `master`, and `schedule` trigger run on default branch (`master`)


server/store/kvstore/poll.go
```diff
@@ -78,6 +80,8 @@ func (s *PollStore) Update(prev *poll.Poll, new *poll.Poll) error {
 
 // Delete deletes a poll from the KV Store.
 func (s *PollStore) Delete(poll *poll.Poll) error {
+	//nolint:revive // Ignore linter error here, because "model.AppError(nil) != nil" always returns true.
```
hanzei, line 83: What is the concrete error? I'm unable to reproduce it locally.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
+name: ci
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  push:
```
hanzei, line 5: If CI doesn't run on `master`, do the commits on `master` get checked? There might be issues because PRs are not always in sync.

hanzei (changes requested):
Thanks for taking care of the migration :+1:

.github/workflows/ci.yml
kaakaa: At that time, I used to think that `master` is checked daily by schedule trigger, so it's not necessary to check when merging PR `master` immediately. But since it's not necessary to remove `master` trigger, I've recovered it.

server/store/kvstore/poll.go
kaakaa: Without this comment, I encountered the following lint error.

```shell
$ go version
go version go1.21.4 linux/amd64
$ golangci-lint version
golangci-lint has version v1.52.2 built with go1.20.2 from (unknown, mod sum: "h1:FrPElUUI5rrHXg1mQ7KxI1MXPAw5lBVskiz7U7a8a1A=") on (unknown)

$ make check-style
...
Running golangci-lint
golangci-lint run ./...
server/store/kvstore/poll.go:83:2: if-return: redundant if ...; err != nil check, just return error instead. (revive)
        if err := s.api.KVDelete(pollPrefix + poll.ID); err != nil {
                return err
        }
make: *** [Makefile:52: check-style] Error 1
```

kaakaa: After updating golangci-lint 1.55.2, the lint error is not detected without comments, so I removed comments to avoid lint errors.

hanzei (approved):

PR #520, opened 2023-11-20
Add missing executables to plugin.json

hangyas:
I tried to run the plugin on arm64, it didn't work, I noticed that there are actually arm64 builds in the server/dist folder, but they are missing from the plugin.json. Patched it and so far it seems to work


kaakaa (approved):
Great catching! Thank you @hangyas 👍

PR #526, opened 2023-12-19
Reduce amount of monkey patching

hanzei:
Monkey patching is a bad practice. We should avoid it as much as possible.

This PR introduces a `poll.Factory` that takes care of the poll creation. In tests, setters can be used to modify the behaviour.

The only test that still needs to get migrated is `TestPluginOnActivate`, which is a PITA to do without dropping coverage. Maybe it's worth removing some test cases if that gets us away from monkey patching.

### DIFF
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -311,7 +311,7 @@ func (p *MatterpollPlugin) handleCreatePoll(_ map[string]string, request *model.
 	userLocalizer := p.bundle.GetUserLocalizer(creatorID)
 
 	settings := poll.NewSettingsFromSubmission(request.Submission)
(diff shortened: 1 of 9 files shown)


hanzei:
> Maybe it's worth removing some test cases if that gets us away from monkey patching.

@kaakaa What do you think? Sometimes I wonder if we went too far with the test coverage for the cost of readability.

kaakaa:
@hanzei My words were not enough. I agree to remove some tests because we are not aiming for 100% coverage and it's difficult to find errors of OnActivate by unit tests.

hanzei:
Let me update the PR

Issue #386, opened 2021-05-31
Anonymous poll authoring


KhaimovMR:
#### Summary

I've seen a lot of cases in the corporate public chats communication that there's a need to have an ability to anonymously create a poll (so the author of the poll will be anonymous).

#### Describe alternatives you've considered

I don't really see the alternative, but I could suggest that there's may be a workaround for such anonymous posting into the channel through the dedicated bot. Am I wrong in this suggestion here?

hanzei:
What is the use case exactly? Is there is reason to hide the author of a poll?

KhaimovMR:
Yes, of course. Thanks for the question.
In a corporate communication culture people quite frequently facing the need to poll their colleagues on some controversial or inconvenient topics and issues.

This kind of communications between colleagues do help to resolve a complex situations without abandoning the subject that keeps disturbing minds without the closures.

Every HR would be very glad to have such instrument in their arsenal (especially in holocracy-like environments).

MichaelKhaimovCW:
Is it a good time investment to make a commit and pull-request on this matter? Or will it be rejected?

KhaimovMR:
Is it a good time investment to make a commit and pull-request on this matter? Or will it be rejected?

KhaimovMR:
@hanzei Is it a good time investment to make a commit and pull-request on this matter? Or will it be rejected?

hanzei:
The use case seems valid to me. I would love if you open a PR for that.

PR #537, opened 2024-03-23
[GH-386] add option anonymous creator

annaos:
- Add new CLI option `--anonymous-creator `
- Add new option to dialog
- Add tests

Issue: #386


hanzei:
@annaos is your branch up to date with `master`?

annaos:
> @annaos is your branch up to date with `master`?

yes, it is up to date

server/poll/transform.go
```diff
@@ -167,6 +171,9 @@ func (p *Poll) makeAdditionalText(bundle *utils.Bundle, numberOfVotes, numberOfV
 	if p.Settings.Anonymous {
 		settingsText = append(settingsText, "anonymous")
 	}
+	if p.Settings.AnonymousCreator {
+		settingsText = append(settingsText, "anonymous-creator")
+	}
```
hanzei, line 176: Unrelated to the PR: Maybe `poll.Settings` needs a `String()` method.

hanzei (approved):
Great work, LGTM :rocket:

hanzei:
The CI is unrelated and will be fixed on the next commit.

server/poll/transform.go
annaos: I did it, but I am unsure, if it is good idea and if it need (so much) tests. At the moment I can found only one place to use this function. It is ok for me to revert last commit :)

kaakaa (changes requested):
Thanks @annaos !

This work looks great, but a creator name still remain in a card, and `Created by` section in a card should also be hidden with `anonymous-creator` setting. (Card can be displayed only if a poll is create with `--progress` setting)

スクリーンショット 2024-03-31 22 28 16

Could you fix this? Related code would be here. https://github.com/matterpoll/matterpoll/blob/master/server/poll/transform.go#L260

kaakaa (approved):
Thanks 👍

Issue #463, opened 2022-11-25
Add support for a single transferable vote


NA-Charles:
#### Summary

Add support for the Single Transferable Vote (STV) system as described in https://en.wikipedia.org/wiki/Single_transferable_vote

#### Additional context

When poll contains more than two choices, voters who chose an unpopular option may have a preference for a "second choice".  The current implementation (sometimes referred to as "First Post the Post") doesn't take this into account.
Using a poll where multiple options can be chosen mitigates this to some extent, but that can bias the results towards common low-ranking options.

#### Possible Implementation

User interaction would only be slightly modified from the current behaviour:
- To specify this voting mechanism, the user could provide an additional parameter (--stv for example)
- This option is mutually exclusive with the --votes=n option
- When voting the user would need to click on their votes in order of preference.
- Clicking on an already selected entry would remove it from the list of votes, clicking on it again would add it to the back of the list.

The interface wouldn't need to change much:
- There should be some indication that STV is active.
- When options are selected, there needs to be an indication of which order they were selected
- The --progress option could remain but there is a question of what you display - do you recalculate the end result each time a vote is cast?
- When the poll ends, the order of winners should be displayed as there may be more than one.

Under-the-hood, there are some more significant functionality changes required:
- The votes cast by each individual need to be recorded, and the order of those votes is important.
- When the "End Poll" option is selected, votes are tallied and the STV system is applied, executing rounds of voting, dropping the least popular each round and transferring the votes allocated to that option, repeating until there is a winner.

PR #539, opened 2024-04-04
[GH-489] add multiple votes for same option

### LINKED ISSUE #489: Same Option Multivote
#### Summary

Currently it is possible to set and distribute several votes, but you can only give **one vote to each option**.
Here it would be nice to have a possibility to place multiple votes on one option.

#### Additional context

The aim is to have a poll in which the most important point for you can be weighted.

For example, if 4 options like "o1" "o2" "o3" "o4" are up for vote and everyone has three (3) votes, multiple votes can be placed on one and same option. 
Like:
```
| Option | "o1" | "o2" | "o3" | "o4" |
+--------+------+------+------+------+
| User-1 |  1   |  1   |  1   |  0   |
| User-2 |  2   |  0   |  1   |  0   |
| User-3 |  0   |  0   |  0   |  3   |

Result:
"o1" = 3 Votes
"o2" = 1 Vote
"o3" = 2 Votes
"o4" = 3 Votes
```

annaos:
Close #489 

- add new setting `--multiple-votes`
- add new field in dialog
- update Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,11 @@ If you want to define all answer options by yourself, type `/poll "Is Matterpoll
 
 Poll Settings provide further customisation, e.g. `/poll "Is Matterpoll great?" "Of course" "In any case" "Definitely" --progress --anonymous`. The available Poll Settings are:
 - `--anonymous`: Don't show who voted for what at the end
+- `--anonymous-creator`: Don't show author of the poll
 - `--progress`: During the poll, show how many votes each answer option got and, in post card, show who voted for which answers ([#431](https://github.com/matterpoll/matterpoll/pull/431))
 - `--public-add-option`: Allow all users to add additional options
 - `--votes=X`: Allow users to vote for X options
+- `--multiple-votes`: Allow users to vote multiple times for the same option
 
 ## Localization
 
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -66,6 +66,10 @@ var (
 		ID:    "command.help.text.pollSetting.public-add-option",
 		Other: "Allow all users to add additional options",
 	}
+	commandHelpTextPollSettingMultipleVotes = &i18n.Message{
+		ID:    "command.help.text.pollSetting.multiple-votes",
+		Other: "Allow users to vote multiple times for the same option",
+	}
 	commandHelpTextPollSettingMultiVote = &i18n.Message{
 		ID:    "command.help.text.pollSetting.multi-vote",
 		Other: "Allow users to vote for X options",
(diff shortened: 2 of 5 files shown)


kaakaa:
Thanks @annaos, but I think the specification is still ambiguous. `--multi-votes` depends on `--votes` setting and the maximum of `--votes` is the number of options, but I don't think it is necessary to follow that limitation for multi-votes. Also, when `--votes=1`, even if `--multi-votes` is set, an user can only vote for one option, which is little weird.

annaos:
Can I do something here? Do you have any ideas?

kaakaa:
It might be just a question of wording, but the existing `--votes=X` setting also has the meaning of "multiple votes," so I don't think `--multiple-votes` is a suitable setting name. As far as I googled, the voting method for #489 is called [Cumulative voting](https://en.wikipedia.org/wiki/Cumulative_voting). (--votes=X would be [Limited voting](https://en.wikipedia.org/wiki/Limited_voting)).

In order to make users aware that the settings we are trying to implement this time are different from current Matterpoll's voting methods, I think it would be better to use one of the following formats for the settings.

1.  Add the setting `--cumulative-votes=X` in addition to `--votes=X` (mutually exclusive)
2. Add a setting such as `--vote-method=cumulative` in addition to `--votes=X`
    * Other possible voting methods include "[approval](https://en.wikipedia.org/wiki/Approval_voting)", "[stf](https://en.wikipedia.org/wiki/Single_transferable_vote)" (#463), and so on.

In both cases, I think it's better to remove the limitation that "the number of votes that can be selected is limited to the maximum number of options" for Cumulative voting.

The above is my thought. I'm glad to hear any thoughts on this, thanks. @annaos @hanzei

annaos:
I like the idea of the` --vote-method`. It opens up more possibilities for the future.  If it is ok with you, I will change this PR to implement this solution.

kaakaa:
Thanks for updating @annaos , and it looks great to me 👍  It may be a work in progress, but I think there are a few things that need to be addressed.

* [`ResetVote`](https://github.com/matterpoll/matterpoll/blob/master/server/poll/poll.go#L261) doesn't support Cumulative correctly yet
* Need to think about [how to validate MaxVotes](https://github.com/matterpoll/matterpoll/blob/master/server/poll/poll.go#L167) in the case of cumulative
  * Should be able to specify any number of votes regardless of the number of vote options, but a limit may be needed (10? 30? 100?)
* Need to tally the votes of the same people in [RHS card](https://github.com/matterpoll/matterpoll/blob/master/server/poll/transform.go#L252) (ref. #431)
* With ExperimentalUI, it would be better to display how many votes an user added to each option. (This can be another issue)

> UPD I tried this solution to have a look at it and did it complete.
> If the solution is ok, I need help with the failed test: I suppose for ... range in server/poll/transform.go:217 change the order of displaNames randomly. How do I get it to work when testing?

Hm... I came up with the following two ways to follow some sort of order, but I prefer 2. in that it can represent the order of votes.

1. Sort keys of `voters` (alphabetical?)
2. Create new slice to keep the order of the voters like the following

```go
...
if !p.Settings.Anonymous {
	voters := make(map[string]int)
	uniques := []string{} // to keep the order of the voters
	for i := 0; i < len(o.Voter); i++ {
		displayName, err := convert(o.Voter[i])
		if err != nil {
			return nil, err
		}
		voters[displayName]++
		if !slices.Contains(uniques, displayName) {
			uniques = append(uniques, displayName)
		}
	}

        for _, u := range uniques {
        ...
...
```

annaos:
@kaakaa I did the first three points and fix the test. Do you want to check it?

kaakaa:
@annaos I'm sorry to my late response, and thanks for your continuous work!

> I did the first three points and fix the test. Do you want to review it?

I'm okay with both. If you still have work in progress after updating the branch, it would be helpful for me to keep it clear.

emes-imtek:
I wanted to highlight something to keep in mind regarding multiple votes and toggling. If `--vote-method=cumulative` isn't set, it's quite easy to reset individual votes f.ex. by simply clicking the same option again. On the other hand, with `--votes=X` and `--vote-method=cumulative` enabled, you'd have to cycle through X clicks to bring a vote count back down to zero. Of course, there’s always the "Reset your votes" option as a fallback.

We’d really love to see `--vote-method=cumulative` make it into a release. We're looking forward to using it for things like placing our lunch orders directly in Mattermost. Thanks for all the hard work!

PR #544, opened 2024-04-14
[GH-95] add end setting to poll

### LINKED ISSUE #95: Feature request: define end date / time of poll
Hi

Would be cool that the user can define when the poll should automatically end.

Perhaps like this --until 01.11.2018 12:00

Best regards

JtheBAB

annaos:
Close #95 

- new setting `--end`
- new field in`Create Poll` dialog
- check and end existing polls by activating the plugin

If this solution is ok, I'll extend testing for new functionality:
- add tests in plugin_test 
- add scheduler_test


kaakaa:
Super @annaos  🚀  I would love to merge this feature. Before that, I have two comments on implementation method.

First, as for the scheduler, instead of implementing our own scheduler, how about using the [`github.com/mattermost/mattermost-plugin-api/cluster.JobOnceScheduler`](https://github.com/mattermost/mattermost-plugin-api/blob/master/cluster/job_once_scheduler.go#L57)? I am not familiar with this API, but it seems that we can let Mattermost manage the jobs, and even if restarting the plugin, Mattermost will restart the jobs without any additional code. With this api, we wouldn't need new store function `GetAllPollIDs`.

Second, I would like to avoid cloning the endPoll code, so it would be great if we could extract code from `handleEndPollConfirm` and make it common func.

kaakaa:
@hanzei I would love to hear from you about using `mattermost-plugin-api/cluster` for this feature.

annaos:
@hanzei I would be happy to hear from you about using `mattermost-plugin-api/cluster` for this feature too.

README.md
```diff
@@ -41,6 +41,8 @@ Poll Settings provide further customisation, e.g. `/poll "Is Matterpoll great?"
 - `--anonymous`: Don't show who voted for what at the end
 - `--progress`: During the poll, show how many votes each answer option got and, in post card, show who voted for which answers ([#431](https://github.com/matterpoll/matterpoll/pull/431))
 - `--public-add-option`: Allow all users to add additional options
+- `--end=2006-01-02T15:04`: End poll automatically at 2006-01-02 at 15:04 or
```
Zeratoxx, line 44: What about the time zone?  
Also mention ISO 8601

annaos: I use the timezone of the author of the poll. I think it is more user-friendly

Zeratoxx: Hmm.. Then this should be mentioned in the documentaion

kaakaa: I agree that it's better to clarify which timezone will be used, and not to handle seconds. (And since [the code parsing datetime string](https://github.com/matterpoll/matterpoll/pull/544/files/6c5b8c5ccf2a2376b48b907c0d072ffda69ee1bf#diff-320751a0c35c5cde04e00dda2b10c4098126fd0486ca09ca85131bfe09607814R183) will be running on Mattermost instance, I think that the timezone of the instance is used, not of the author. Is my understanding correct?)

annaos: I am almost back from holiday :blush: @kaakaa You are write, the timezone of the instance is used.

- I updated the help text
- Added more ways to parse the date (pass with or without seconds and timezone)
- Moved some logic from poll.go to new setting.go file

I think I did everything I wanted and PR is ready for review

kaakaa (changes requested):
@annaos Thanks again, I really appreciate it!
I left a lot of comments for bug fixes, improvements, and suggestions, so I'd appreciate it if you could take a look.

PR #564, opened 2024-09-15
Set omitempty to AnonymousCreator setting

### LINKED ISSUE #562: Cannot update old poll after updating to v1.7.0
#### Summary
Cannot update old poll after updating to v1.7.0

#### Steps to reproduce
1. Create poll use v1.5.0.
2. Update plugin to v1.7.0.
3. Any update action like vote, add-option and so on 

#### Expected behavior
Action success.

Action failure

image

#### Server configuration

**Mattermost version:** Mattermost Enterprise Edition v9.5.7

**Matterpoll version:** v1.7.0

**Updated from an older Matterpoll version or fresh install:** update from v1.5.0

<details>

```
{"timestamp":"2024-09-10 17:11:19.497 +08:00","level":"warn","msg":"failed to handle PostActionIntegrationRequest","caller":"app/plugin_api.go:1006","plugin_id":"com.github.matterpoll.matterpoll","error":"failed to save poll: poll already exists in database"}
```
</details>

#### Additional context
I think it might be related to [this](https://github.com/matterpoll/matterpoll/pull/537)(data integration issue?).
can you confirm this?

kaakaa:
fix #562 

## Summary
Since the poll craeted by Matterpoll v1.6 or before doens't have `AnonymousCreator` setting, atomic transaction will be failed. Set omitempty to `AnonymousCreator` to avoid this error.

After merging this, get ready for releasing v1.7.1.

---

An upgrade script to update data within the KV Store could be considered, but it would need to be run every time a release that has new field for Poll/Settings. Therefore, a simpler approach using omitempty was adopted. The upgrade script will only be used when it is necessary to set specific initial values.

### DIFF
--- a/server/poll/poll.go
+++ b/server/poll/poll.go
@@ -23,6 +23,8 @@ const (
 )
 
 // Poll stores all needed information for a poll
+// When adding new fields, to avoid failures during atomic transactions for KV Store,
+// either specify omitempty or set initial values during the upgrade.
 type Poll struct {
 	ID            string
 	PostID        string `json:"post_id,omitempty"`
@@ -39,10 +41,12 @@ type AnswerOption struct {
 	Voter  []string
 }
 
-// Settings stores possible settings for a poll
+// Settings stores possible settings for a poll.
+// When adding new settings, to avoid failures during atomic transactions for KV Store,
+// either specify omitempty or set initial values during the upgrade.
 type Settings struct {
 	Anonymous        bool
-	AnonymousCreator bool `json:",omitempty"`
+	AnonymousCreator bool `json:",omitempty"` // Since v1.7
 	Progress         bool
 	PublicAddOption  bool
 	MaxVotes         int `json:"max_votes"`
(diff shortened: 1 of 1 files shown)


hanzei (approved):
Thanks for adding the comments :+1:

server/poll/poll.go
```diff
@@ -39,10 +41,12 @@ type AnswerOption struct {
...
 }
 
-// Settings stores possible settings for a poll
+// Settings stores possible settings for a poll.
+// When adding new settings, to avoid failures during atomic transactions for KV Store,
+// either specify omitempty or set initial values during the upgrade.
 type Settings struct {
 	Anonymous        bool
-	AnonymousCreator bool
+	AnonymousCreator bool `json:",omitempty"`
```
hanzei, line 49: Should we also specify the JSON field name here?

hanzei (commented):
Can we add a test that covers the fix?

server/poll/poll.go
kaakaa: Thanks @hanzei , but I'll leave it as is for now.  Because if we specify the field name, it must be `json:"AnonymousCreator,omitempty"` for compatibility, but it looks little weird.  And features works fine without specifying the JSON field name.

